PR #1, opened 2016-11-14
Implement rosca interface

shine2lay:
there might bugs or things that I missed , pls review and let me know


contracts/rosca.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity ^0.4.4;
 /*
 *  Possible missing functions
 * - integrity check? to perform check and make sure certain values are available and correct
-* TODO: a way to randomly pick a member that havent got paid this
+*
 */
 contract rosca {
```
rmerom: should we follow the Solidity style guide and name contracts in CapWords?
http://solidity.readthedocs.io/en/develop/style-guide.html#contract-and-library-names

contracts/rosca.sol
```diff
@@ -70,8 +72,8 @@ contract rosca {
      */
     function rosca(
```
rmerom: do you think we can have comments above the contract and the methods? it can help with future code maintenance, and we'll need to add them anyway before security audit, ICO etc.

contracts/rosca.sol
```diff
@@ -80,8 +82,8 @@ contract rosca {
         roundPeriodInDays = _roundPeriodInDays;
         if(roundSum < 1) throw; //if everyone is only required to contribute 0 , whats the point of rosca
```
rmerom: suggestion: have a constant MIN_ROUND_SUM = 1000000000000; (1e+12 wei == 1e-6 ether ~~ 1e-7 USD today). I think this,or any other reasonable number that makes sense to you, could prevent human errors.

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
-    event LogContributionMade(string uniqName);
-    event LogBidPlaced(string uniqName);
-    event LogRoundFundsRelease(string winnerUniqName, uint amountInWei);
-    
-    
-    enum Period { Weekly, Monthly }
-    
+    address constant WETRUST = 0x0;
+
+    enum Period { Weekly, Monthly } // are we only allwing 2 options?
```
rmerom: We should probably either use this, or remove it from the code.

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+
+    event LogParticipantRegistered(address user);
+    event LogContributionMade(address user, uint amount);
+    event LogNewLowestBid(uint bid,address winnerAddress);
+    event LogRoundFundsRelease(address winnerUniqName, uint amountInWei);
+
+
+    //added
```
rmerom: thanks for the tip, but i think we don't this anymore

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
-    enum Period { Weekly, Monthly }
-    
+    address constant WETRUST = 0x0;
+
+    enum Period { Weekly, Monthly } // are we only allwing 2 options?
+
+    event LogParticipantRegistered(address user);
+    event LogContributionMade(address user, uint amount);
+    event LogNewLowestBid(uint bid,address winnerAddress);
+    event LogRoundFundsRelease(address winnerUniqName, uint amountInWei);
```
rmerom: s/winnerUniqName/winnerAddress

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    event LogFundsWithdrawal(address user, uint amount,address destination);
+    event LogStartOfPeriod(uint currentRound);
+
+    //state variables
+    address foreman;
+    uint roundSum;
+    uint numRounds;
+    uint currentEpoch;
+    uint numEpochs;
+    int8 minParticipants;
```
rmerom: nit: just to prevent small bugs if someone decides to run a bigger ROSCA, perhaps we can change this to int16? anyway this won't use up more space because the EVM bunches together variables to 256-bit chunks

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    uint roundPeriodInDays;
+    uint startTime;
+    int16 feeToService;
```
rmerom: You might want to put together the non-256-bit vars, so that EVM will crunch them into one 256-bit word.
http://solidity.readthedocs.io/en/develop/miscellaneous.html#layout-of-state-variables-in-storage

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+
+
+    //added
+    event LogFundsWithdrawal(address user, uint amount,address destination);
+    event LogStartOfPeriod(uint currentRound);
+
+    //state variables
+    address foreman;
+    uint roundSum;
+    uint numRounds;
```
rmerom: nit: the round and epoch counters can probably be reduced to uint16

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    uint numEpochs;
+    int8 minParticipants;
+    uint roundPeriodInDays;
+    uint startTime;
+    int16 feeToService;
+    address feeAddress;
+    uint currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
+
+    /*
+        not sure about which structures to use yet
```
rmerom: does this still hold?

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    */
+    struct USER{
+        uint contributed; // number of times contributed or could be amount contributed
+        uint paid; // number of times bid Won
+        uint allowance; // how much they are allowed to withdraw, i.e if someone win , their allowance will go up by the bid.
+        bool alive; //needed to check if a member is indeed a member
+
+    }
+    mapping(address => USER) members; //using struct USER to keep track of contributions and paid, allowance and etc.
+    address[] memberAddress;    //this is the only way to iterate through all the member's address
```
rmerom: nit: 
s/member's address/members' addresses

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+
+    mapping(address => bool) pendingJoinRequest; // this way , address can be used as index, if we use address[] , we'll have to go thru a whole array
+
+    //deposit and withdrawal related
+    uint totalAmountInContract;
```
rmerom: do we need this? shouldn't this.value hold the same value?

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    //bidding related state variable
+    uint lowestBid;
+    address winnerAddress;
+
+
     modifier onlyForeman {
+        if(msg.sender != foreman)
+        throw;
```
rmerom: should either be on same line, or indented; same on Line 65

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        _;
+    }
+    modifier beforeStart {
+        if(currentRound == 0)
```
rmerom: if this modifier only applies before starting, shouldn't this be

if (currentRound != 0) throw;

?

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
-        uint roundSum, 
-        int16 numRounds, 
-        int8 minParticipants, 
-        int startTime, 
-        int16 feeInThousandths, 
-        address feeAddress);
-    
+    function rosca(
+        uint _roundPeriodInDays,
+        uint _roundSum, //i am assuming this is referring to number epochs to run
```
rmerom: I would think _numEpochs on same line refers to # of epochs to run, whereas this is the contribution per round?

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        uint _numEpochs,
+        int8 _minParticipants,
+        uint _startTime,
+        int16 _feeInThousandths,
+        address _feeAddress)
```
rmerom: I don't remember exactly what we discussed back in Starbucks - shouldn't this be a constant?

contracts/rosca.sol
shine2lay: when i was writing this contract, i had "ether" in mind instead of "wei" for units, I think wei is a better unit to use

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i just realized it as i was going thru the code just now, i already removed it

contracts/rosca.sol
shine2lay: i noticed this just now as well, fixed

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    {
+        roundPeriodInDays = _roundPeriodInDays;
+        if(roundSum < 1) throw; //if everyone is only required to contribute 0 , whats the point of rosca
+        roundSum = _roundSum;
+        if(_numEpochs < 1) throw;
+        numEpochs = _numEpochs;
+        if(_minParticipants < 2) throw;// there should be at least 2 people to make a group
+        minParticipants = _minParticipants;
+        if(_startTime < now) throw;
```
rmerom: nit: might want to add a minimum period for starting a ROSCA, e.g. 1 day or something (so one can't start a ROSCA in 5 minutes)

contracts/rosca.sol
shine2lay: yes, i'll change it to contributionSize to avoid confusion

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        startTime = _startTime;
+        if(_feeInThousandths < 0) throw; //fee must be non-negative
```
rmerom: Please also check against a MAX_FEE_IN_THOUSANDTH = XXX;
e.g. XX might be 20 (2%).

contracts/rosca.sol
rmerom: nit: maybe call this serviceFeeInThousandths?

contracts/rosca.sol
shine2lay: from what i remember, you said, by default fees would be like 0.3% but the user can specify the fee if they decide to pay more or less. i think the feeAddress should be constant.

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        feeToService = _feeInThousandths;
+        if(feeAddress != 0) //feeAddress shouldnt be empty(null)
+        feeAddress = _feeAddress;
+
+
```
rmerom: Please add
foreman = msg.sender;

shine2lay: done

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    address feeAddress;
+    uint currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
+
+    /*
+        not sure about which structures to use yet
+    */
+    struct USER{
+        uint contributed; // number of times contributed or could be amount contributed
+        uint paid; // number of times bid Won
+        uint allowance; // how much they are allowed to withdraw, i.e if someone win , their allowance will go up by the bid.
```
rmerom: nit: in some other contracts, I saw this named pendingWithdrawl . We might want to use the same name (as allowance has some other meanings too)

shine2lay: agreed, done

contracts/rosca.sol
rmerom: is this the current round number? if so, consider naming something like roundNum, as numRounds may be understood as "number of rounds"

rmerom: I actually see that there is also a currentRound, so I'm not sure what roundNum means?

shine2lay: numRounds is number of Rounds in one epoch

shine2lay: i added explanation with comment next to it

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    function rosca(
+        uint _roundPeriodInDays,
+        uint _roundSum, //i am assuming this is referring to number epochs to run
+        uint _numEpochs,
+        int8 _minParticipants,
+        uint _startTime,
+        int16 _feeInThousandths,
+        address _feeAddress)
+    {
+        roundPeriodInDays = _roundPeriodInDays;
```
rmerom: nittiest nit: I slightly prefer we use 2-spaces indentation, as we can fit more code then. However, if you prefer 4-spaces, that's completely fine by me.

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        if(_feeInThousandths < 0) throw; //fee must be non-negative
+        feeToService = _feeInThousandths;
+        if(feeAddress != 0) //feeAddress shouldnt be empty(null)
+        feeAddress = _feeAddress;
+
+
+        //register foreman as a member of ROSCA and increment numRounds in a Epoch to 1
+        members[msg.sender] = USER({paid: 0 , contributed: 0, alive: true, allowance: 0});
+        memberAddress.push(msg.sender);
+        numRounds = 1;
```
rmerom: (relating to a comment above) I'm not sure what numRounds is, compared to currentRound.

shine2lay: numRounds in one epoch

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    }
+
+
+
+    function startRound()
```
rmerom: Is this an external function?
if so, who is responsible to calling it?
if not, consider adding the "internal" modifier to it.

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: agree, WIP

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+    {
+
+        if(now < startTime + ( currentRound * currentEpoch  * (roundPeriodInDays * 1 days)))
+            throw;
+        else
```
rmerom: no need for "else" (we throw if the "if" applies)

contracts/rosca.sol
shine2lay: i believe we mentioned this during our meeting, at the end of each round, we need to call a function to get the winningBid and winningAddress otherwise there is no way for the contract to know when to end the round. Unless we use services like "Alarm clock"

contracts/rosca.sol
```diff
@@ -1,72 +1,266 @@
...
+        memberAddress.push(msg.sender);
+        numRounds = 1;
+    }
+
+
+
+    function startRound()
+    {
+
+        if(now < startTime + ( currentRound * currentEpoch  * (roundPeriodInDays * 1 days)))
```
rmerom: shouldn't this be startTime + ( (currentEpoch \* numRoundsPerEpoch) + currentRound) \* roundPeriodInDays \* 1 days
?

rmerom (changes requested):
Hey Shine, looks good!
I reviewed till somewhere in the middle. Let's get some of the comments addressed, then continue.

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: wouldnt uint8 or uint16 would be better? since they shouldnt be negative

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i think the struct is a good idea, do you have a better idea?

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: can you clarify a bit more?

contracts/rosca.sol
shine2lay: ok will look into it , WIP

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i changed it to uint16 for now, because i think its a better choice but we can discuss more on this later

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: changed it to membersAddresses

contracts/rosca.sol
shine2lay: added now + 1 days

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: yes, you are right , fixed it

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: nope, i took it out

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i dont mind either way, i'll try to set everything to 2 space indentation

contracts/rosca.sol
rmerom: sorry, it's an abbreviation i use a lot, i should have checked... :)

s/something/different means: change "something" to "different".
(it's coming from vi commands: http://vim.wikia.com/wiki/Search_and_replace

contracts/rosca.sol
rmerom: you're right, uint16 IS better, thanks!

contracts/rosca.sol
rmerom: great, that's what i meant too :)

contracts/rosca.sol
rmerom: thanks i appreciate it!

contracts/rosca.sol
rmerom: I see. That's a good point.
However, I'm not sure who will call this. We may need to notify the foreman to call this function, but it's a bit of a hassle for him/her. 
Do you think whenever someone calls withdraw() or bid() we can check whether a new round has begun already?

rmerom (changes requested):
Great work!
I'm going to review the new PR now

PR #2, opened 2016-11-15
Implement rosca interface

shine2lay:
i did the changes you requested, let me know if i missed somethings


contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
 pragma solidity ^0.4.4;
 
-/*
-*  Possible missing functions
-* - integrity check? to perform check and make sure certain values are available and correct
-*
+/**
+*  Things still missing/needs fixing:
+*   in bid(), check if the ROSCA/rounds is started before allow bidding
+*   in startRound() , check if minimumParticipants requirement is met (what happens if it doesn't? should we move the startTime?)
```
rmerom: For MVP, I think we can have the ROSCA fail if not enough participants are there.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-    uint roundPeriodInDays;
-    uint startTime;
-    int16 feeToService;
-    address feeAddress;
-    uint currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
-
-    /*
-        not sure about which structures to use yet
+contract ROSCA {
+  address constant WETRUST = 0x0;
```
rmerom: s/WETRUST/WETRUST_FEE_ADDRESS

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  uint constant MIN_CONTRIBUTION_SIZE = 1000000000000;
```
rmerom: consider adding 
// 1e12 wei

at end of line

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    if(contributionSize_ < MIN_CONTRIBUTION_SIZE) throw; //if everyone is only required to contribute 0 , whats the point of rosca
+    contributionSize = contributionSize_;
+
+    if(numEpochs_ < 1) throw;
+    numEpochs = numEpochs_;
+
+    if(minParticipants_ < 2) throw;// there should be at least 2 people to make a group
+    minParticipants = minParticipants_;
+
+    if(startTime_ < (now + 1 days)) throw;
```
rmerom: I suggest we try to avoid "magic numbers" in the code like "1 days" - making them constants should not change the code output size

http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-    int16 feeToService;
-    address feeAddress;
-    uint currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
-
-    /*
-        not sure about which structures to use yet
+contract ROSCA {
+  address constant WETRUST = 0x0;
+  uint constant MIN_CONTRIBUTION_SIZE = 1000000000000;
+  uint constant MAX_FEE_IN_THOUSANDTH = 20;
```
rmerom: 20 thousandths is 0.2%. As far as I remember, @cottonbrew mentioned 0.1%-1% fee. Can we make this 200 thousandths to be on the safe side?

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+
+  event LogParticipantRegistered(address user);
```
rmerom: We may need to break this to 2 events:
1. LogParticipantApplied(address user)
for when a user wants to join a ROSCA, and
1. LogParticipantApprove(address user)
   for when a user was approved by the foreman

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  event LogContributionMade(address user, uint amount);
+  event LogNewLowestBid(uint bid,address winnerAddress);
+  event LogRoundFundsRelease(address winnerUniqName, uint amountInWei);
```
rmerom: 1. s/LogRoundFundsRelease/LogRoundFundsReleased
2. s/winnerUniqName/winnerAddress (because we no longer have unique name strings for participant)

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  event LogFundsWithdrawal(address user, uint amount,address destination);
+  event LogStartOfPeriod(uint currentRound);
+
+  //state variables
```
rmerom: nitpick: space after // according to style guide (here and in other lines), e.g.
// state variables

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  uint16 numRoundsInEpoch;     //number of rounds in 1 epoch, user doesnt have control over this, it'll increment everytime a new participant is registered
+  uint16 currentEpoch;
```
rmerom: i think it's great that you wrote code for the entire ROSCA, but since it's a lot of code, there might be security issues that will be harder to find. I think it might make sense to start with 1 epoch per ROSCA for simplicity, make sure it works correctly, then add the ability to run more than one epoch. What do you think?

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: couldn't just set it as a parameter?

shine2lay: couldn't we just set it as a parameter while testing?

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+
+  event LogParticipantRegistered(address user);
+  event LogContributionMade(address user, uint amount);
+  event LogNewLowestBid(uint bid,address winnerAddress);
+  event LogRoundFundsRelease(address winnerUniqName, uint amountInWei);
+  event LogFundsWithdrawal(address user, uint amount,address destination);
+  event LogStartOfPeriod(uint currentRound);
+
+  //state variables
+  uint16 numRoundsInEpoch;     //number of rounds in 1 epoch, user doesnt have control over this, it'll increment everytime a new participant is registered
```
rmerom: If I'm not mistaken, it always holds that numRoundsInEpoch == membersAddresses.length . Can we use that to save both space and instructions?

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  uint16 currentEpoch;
+  uint16 numEpochs;     //number of Epochs this contract will run
+  uint16 roundPeriodInDays;
+  uint16 serviceFeeInThousandths;
+  uint16 currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
```
rmerom: nit: s/start/starts

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  uint8 minParticipants;
```
rmerom: thanks for bunching together the smaller vars! :)

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: np

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  address foreman;    //address of the foreman
+  address feeAddress;
+  uint contributionSize;
+  uint startTime;     //startTime must be in timestamp
+
+  struct USER{
+    uint contributed; // number of times contributed or could be amount contributed
```
rmerom: if we're not sure yet about the meaning of a var and not using it yet, i suggest removing it altogether. The reasoning is that we might leave unused variables in the code if we commit code this way. Not a big thing, you can decide.

contracts/rosca.sol
shine2lay: yes, done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    uint paid; // number of times bid Won
+    uint pendingWithdrawl; // how much they are allowed to withdraw, i.e if someone win , their allowance will go up by the bid.
```
rmerom: nit: 
s/win/wins
s/allowance/pendingWithdrawl

contracts/rosca.sol
shine2lay: contributed is used when people deposit money

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  uint16 roundPeriodInDays;
+  uint16 serviceFeeInThousandths;
+  uint16 currentRound;  //currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually start
+  uint8 minParticipants;
+  address foreman;    //address of the foreman
+  address feeAddress;
+  uint contributionSize;
+  uint startTime;     //startTime must be in timestamp
+
+  struct USER{
```
rmerom: I think according to the style guide, struct names are CapWords, e.g. User, for example in:
http://solidity.readthedocs.io/en/develop/style-guide.html#control-structures

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  address[] membersAddresses;    //this is the only way to iterate through all the member's address
+
+  mapping(address => bool) pendingJoinRequest; // this way , address can be used as index, if we use address[] , we'll have to go thru a whole array
+
+  //bidding related state variable
+  uint lowestBid;
+  address winnerAddress;
+
+
+  modifier onlyForeman {if(msg.sender != foreman) throw;_;}
```
rmerom: consider breaking into lines, as there are multiple statements inside the braces and it might be easy to miss a mistake:
modifier onlyForeman {
  if (msg.sender != foreman) throw;  // if there is just one statement, can leave on same line.
  _;
}

Same below for beforeStart

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  modifier beforeStart {if(currentRound != 0)throw;_;}
```
rmerom: consider s/beforeStart/onlyBeforeStart

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+
+  /**
+    * Creates a new ROSCA and initialize the necessary variables, ROSCA doesnt start untill the specified startTime
```
rmerom: just some nits: 
s/initialize/initializes
s/untill/until

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    * Creator of the contract becomes foreman and also added as the first member of the ROSCA
+    * params:
+    * - roundPeriodInDays uint16
```
rmerom: i think there is a way to specify param explanations in a structured way , so that tools can later use that:
https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format

Even otherwise, I suggest not mentioning the param type in the comments, as it may get out of sync with the actual param (when someone changes the type of the actual param)

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    * - contributionSize uint //In Wei, amount that every member of the ROSCA is required contribute each period
+    * - numEpoch uint16
+    * - startTime uint // must be as a timeStamp
+    * - feeInThousandths uint16, //fee to the service
+    * - address feeAddress, // a WeTrust owned address to send the fees.
```
rmerom: I think we don't need feeAddress, we have the WETRUST (or WETRUST_FEE_ADDRESS) constant above

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-
-    function startRound()
-    {
-
-        if(now < startTime + ( currentRound * currentEpoch  * (roundPeriodInDays * 1 days)))
-            throw;
-        else
+  function ROSCA(
+    uint16 roundPeriodInDays_,
+    uint contributionSize_, //i am assuming this is referring to number epochs to run
```
rmerom: i think this comment should be on the next line.
as mentioned above, maybe we should start with just 1 epoch always?

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    uint16 numEpochs_,
+    uint8 minParticipants_,
+    uint startTime_,
+    uint16 feeInThousandths_,
+    address feeAddress_)
+  {
+    roundPeriodInDays = roundPeriodInDays_;
```
rmerom: this needs checking, e.g.
if (roundPeriodInDays < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodIndays > MAX_ROUND_PERIOD_IN_DAYS) throw;

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    if(contributionSize_ < MIN_CONTRIBUTION_SIZE) throw; //if everyone is only required to contribute 0 , whats the point of rosca
```
rmerom: I think the comment may not be relevant anymore

contracts/rosca.sol
shine2lay: took out params part

contracts/rosca.sol
shine2lay: already taken care of when i added WETRUST_FEE_ADDRESS

contracts/rosca.sol
shine2lay: removed

contracts/rosca.sol
shine2lay: removed

contracts/rosca.sol
shine2lay: added

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    //register foreman as a member of ROSCA and increment numRounds in a Epoch to 1
+    members[msg.sender] = USER({paid: 0 , contributed: 0, alive: true, pendingWithdrawl: 0});
+    membersAddresses.push(msg.sender);
+    numRoundsInEpoch = 1;
+  }
+
+  /** startRound() check if the round has ended, if so, get the winner of the currentRound's pot
+    * and add the amount to the winner's pending withdrawal
+    * if there were no bid during the round, address at index "currentRoundNumber" is chosen,
+    * if the address chosen already won the pot once in currentEpoch, currentRoundNumber+1 is chosen, untill a winner is found
```
rmerom: suggestion: to make the ROSCA more fun, what if we select a semi-random winner in case there is no bid?
for example, we can start from (block.blockhash(block.number) % [numParticipants])

it's just a nice to have...

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    *
+    * after the winner had been declared, bidding process reset and currentRound is incremented, if currentRound+1 is equal numRoundsInEpoch,
```
rmerom: It's great that you explain what the function does.
However, we may want to be less specific in the documentation, otherwise it starts duplicating the code and will eventually fall out of sync.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    * currentEpoch is incremented, and currentRound = 1;
+    *
+    * if currentRound = 0(ROSCA hasnt started), no winner is declared
+    */
+  function startRound()
+  {
+
+    if(now < startTime + (uint(currentRound + (currentEpoch * numRoundsInEpoch))  * (uint(roundPeriodInDays) * 1 days)))
```
rmerom: wow! this is one reason I thought we may want to avoid epochs for now, as there might be a bug here (even converting between var types) which we'll have a hard time finding (especially as we both are new to Solidity).

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-            {
-                currentEpoch++;
-                currentRound = 1;
-                lowestBid = roundSum;
-
-            }
-
-
+          if(members[membersAddresses[(currentRound-1 + i) % membersAddresses.length]].paid < currentEpoch)
+          winnerAddress = membersAddresses[currentRound-1 + i];
```
rmerom: don't we need a 
break;

after this line?

contracts/rosca.sol
shine2lay: i added a constant for numEpochs value

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i used "now" timestamp as a source of randomness

contracts/rosca.sol
rmerom: nit: according to the style guide, there should be a space after if, for, etc., e.g.:
if (foo) {
  bar;
}
http://solidity.readthedocs.io/en/develop/style-guide.html#control-structures

contracts/rosca.sol
shine2lay: yes, i added it

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    * currentEpoch is incremented, and currentRound = 1;
+    *
+    * if currentRound = 0(ROSCA hasnt started), no winner is declared
+    */
+  function startRound()
+  {
+
+    if(now < startTime + (uint(currentRound + (currentEpoch * numRoundsInEpoch))  * (uint(roundPeriodInDays) * 1 days)))
+      throw;
+    if(currentRound != 0 ){
```
rmerom: nit: remove space before ), add after )

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+      if(winnerAddress == 0)
```
rmerom: can you add a comment about when we expect winnerAddress == 0 to be true vs. when not?

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-     */
-    function joinRequest() beforeStart
-    {
-        //only put the request in the pending list if they are not in the ROSCA already
-        if(!members[msg.sender].alive)
-            pendingJoinRequest[msg.sender] = true;
-        else
-            throw;
+      }
+      members[winnerAddress].pendingWithdrawl += lowestBid;
```
rmerom: The whole logic seems right to me, but as it's pretty complex with many if's and the stakes are high (this is real money), I'm a bit concerned abut this line - what if someone somehow manages to get to it twice? then they'll be "stealing" the win twice.
Can we, just for sake of security, double check and throw if needed, members[winnerAddress].paid before giving them money? 
I think it may be worth the extra gas.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  }
+  /**
+   * Anyone not already a member of ROSCA can request to join and they'll be put into
+   * pendingJoinRequest untill foreman accept request or ROSCA has started
+   *
+   */
+  function joinRequest() beforeStart
+  {
+    //only put the request in the pending list if they are not in the ROSCA already
+    if(!members[msg.sender].alive)
```
rmerom: consider removing the throw - if someone asks to join twice, it's not such a bad behavior, and by throwing, we penalize them by taking all available gas.
If you decide to leave the throw intact, I suggest changing to

if (!a) throw;
doSomething()

as it may be easier to read.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+      pendingJoinRequest[msg.sender] = true;
+    else
+      throw;
+  }
+  /**
+   * foreman must pass in the address of the requtor to approve the ROSCA participation
```
rmerom: it's good to comment, but this may redundant information as anyone who reads the method signature will likely understand that.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-                return true;
-            }
-        }
-        else
-            throw;
-
-    }
-
-    function integrityCheck(uint option) // option will specify whether it will be a message call or transaction call to freeze the contract
+    else throw;
```
rmerom: same logic applies here as above - we may not want to throw() because it'll just waste the foreman's gas

contracts/rosca.sol
shine2lay: that line check if they are already registered as members , not check if they already put in a request

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  }
+
+  // v2: function leave();
```
rmerom: Suggestion: Let's remove dead code to ease reading. When we implement this, we can bring it back.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-            winnerAddress = msg.sender;
-            LogNewLowestBid(lowestBid, winnerAddress);
-
-        }else
-            throw;
+      members[requestor] = USER({paid: 0 , contributed: 0, alive: true, pendingWithdrawl: 0});
+      membersAddresses.push(requestor);
+      numRoundsInEpoch++;
+      LogParticipantRegistered(requestor);
+      delete(pendingJoinRequest[requestor]); // take out the requestor's address in the pending list
```
rmerom: Cool! I actually forgot we need to do this :)

Once ROSCA starts, we may want to delete all the remaining requests (those that were not approved) - it'll both give a gas refund to the caller, and will be easier to debug

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+   */
+  function joinRequest() beforeStart
+  {
+    //only put the request in the pending list if they are not in the ROSCA already
+    if(!members[msg.sender].alive)
+      pendingJoinRequest[msg.sender] = true;
+    else
+      throw;
+  }
+  /**
```
rmerom: Please add a blank line between functions:
http://solidity.readthedocs.io/en/develop/style-guide.html#blank-lines

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  function contribute() payable
+  {
+    //Product condsideration : are we gonna limit how much ether could be sent? is it total, or how much sent each time?
+    if(members[msg.sender].alive)
     {
-        /* Checklist for integrityCheck
-        *
-        */
-
+      members[msg.sender].contributed = msg.value;
```
rmerom: if i understand correctly:
if a participant mistakenly contributes twice in the same round, the money will be stuck there...
consider changing to += or finding another solution

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+
+  // v2: function leave();
+
+  /**
+   * Processes a periodic contribution from msg.sender ().
+   * Any excess funds will be withdrawable through withdraw().
+   */
+  function contribute() payable
+  {
+    //Product condsideration : are we gonna limit how much ether could be sent? is it total, or how much sent each time?
```
rmerom: We can throw if a user sends too much, but I prefer we just credit them (e.g. put it in pendingWithdrawls) - this will allow users to e.g. contribute 2 rounds' worth of money at one time, if they prefer.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
-
-    /**
-     * Vouches that uniqName is now using newWallet as its primary account.
-     * If X-out-of-Y people vouch, the address is updated after a timelock,
-     * unless cancelRestoreAccess() is called from the old address.
-     */
-    function restoreAccess(string uniqName, address newWallet){
-        throw; //for now
+    else
+      throw;
```
rmerom: similar to above - consider putting 

if (!members[msg.sender].alive) throw;

in the first line so the code is localized (i.e. in L195 one does not have to look back up at L188), and there are less indentations and blocks in the code.

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  function bid(uint distrubtionAmountInWei)
+  {
+    if(distrubtionAmountInWei < lowestBid || !(members[msg.sender].paid < currentEpoch))
+    {
+      //set new lowestBid and winnerAddress, also trigger event new Bid placed
+      lowestBid = distrubtionAmountInWei;
+      winnerAddress = msg.sender;
+      LogNewLowestBid(lowestBid, winnerAddress);
+    }else
+      throw;
```
rmerom: same comments as above about throw (prefer not to, to save gas. if must, consider doing so in first line).

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+  }
+
+  /**
+   * Registers a bid from msg.sender. If msg.sender has already won a round or bid is higher than lowestBid,
+   * this method will throw.
+   */
+  function bid(uint distrubtionAmountInWei)
+  {
+    if(distrubtionAmountInWei < lowestBid || !(members[msg.sender].paid < currentEpoch))
```
rmerom: 1. we might want to check that the ROSCA has started, and that a round is in progress.
2. i think the way it's written right now, if i haven't yet won, but my bid is higher than the best bid, i still win? maybe this should be && instead of || ?

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+    {
+      //set new lowestBid and winnerAddress, also trigger event new Bid placed
```
rmerom: nit: in implementation comments i think it's best to explain _why_ the code does it does, or _how_, but not _what_ it does - it's both evident from the code, and causes duplication that can fall out of sync later

contracts/rosca.sol
rmerom: i'm thinking maybe we should have a MIN_DISTRIBUTION_RATIO so that no one, even accidentally, loses all their rounds' money because they, say, mistakenly sent 1 wei instead of 1 ether.
In real life, Chit Funds have that and in their case it's 0.65:
https://www.quora.com/Personal-Savings-How-do-chit-funds-work

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+   * sends the fund to that address.
+   */
+  function withdraw(address opt_destination)
+    returns(bool success)
+  {
+    if(members[msg.sender].alive && members[msg.sender].pendingWithdrawl > 0)
+    {
+      uint amountToWithdraw = members[msg.sender].pendingWithdrawl;  //use a temporary variable to avoid the receiver calling the withdraw function again before
+      members[msg.sender].pendingWithdrawl = 0;                      // send() completes its task
+      if (!opt_destination.send(amountToWithdraw)) {   //if the send() fails, put the allowance back to its original place
```
rmerom: beautiful! :)

contracts/rosca.sol
```diff
@@ -1,266 +1,258 @@
...
+   * Withdraws available funds for msg.sender. If opt_destination is specified,
+   * sends the fund to that address.
+   */
+  function withdraw(address opt_destination)
+    returns(bool success)
+  {
+    if(members[msg.sender].alive && members[msg.sender].pendingWithdrawl > 0)
+    {
+      uint amountToWithdraw = members[msg.sender].pendingWithdrawl;  //use a temporary variable to avoid the receiver calling the withdraw function again before
+      members[msg.sender].pendingWithdrawl = 0;                      // send() completes its task
```
rmerom: since opt_desitination is optional, i think we need to add:
if (opt_destination == 0)
  opt_destination = msg.sender;

rmerom (commented):
Great and fast work, Shine!

contracts/ROSCA.sol
```diff
@@ -0,0 +1,224 @@
+pragma solidity ^0.4.4;
+
+/**
+*  Things still missing/needs fixing:
+*   in withdraw() , add fee related logic
+*   checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+*/
+contract ROSCA {
+  address constant WETRUST = 0x0;
```
rmerom: didn't we name this WETRUST_FEE_ADDRESS ?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,224 @@
...
+  uint constant MIN_CONTRIBUTION_SIZE = 1000000000000;  // 1e12
+  uint constant MAX_FEE_IN_THOUSANDTH = 200;
```
rmerom: s/THOUSANDTH/THOUSANDTHS

contracts/ROSCA.sol
```diff
@@ -0,0 +1,224 @@
...
+  address constant WETRUST_FEE_ADDRESS = 0x0;
+  uint constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;
+  uint constant MINIMUM_PARTICIPANTS = 2;
+  uint16 constant EPOCH_TO_RUN = 1;
+  uint constant MIN_ROUND_PERIOD_IN_DAYS = 1;
+  uint constant MAX_ROUND_PERIOD_IN_DAYS = 30;
+  uint constant MIN_DISTRIBUTION_RATIO = 65;
+
+  event LogParticipantApplied(address user);
+  event LogParticipantApprove(address user);
```
rmerom: My mistake in my previous comment: 
should be LogParticipantApprove_d_

contracts/ROSCA.sol
```diff
@@ -0,0 +1,224 @@
...
+  event LogContributionMade(address user, uint amount);
+  event LogNewLowestBid(uint bid,address winnerAddress);
+  event LogRoundFundsReleased(address winnerAddress, uint amountInWei);
+  event LogFundsWithdrawal(address user, uint amount,address destination);
+  event LogStartOfPeriod(uint currentRound);
+
+  // state variables
+  //uint16 numRoundsInEpoch;     // number of rounds in 1 epoch, user doesnt have control over this, it'll increment everytime a new participant is registered
```
rmerom: nit: removing dead code can help with understandability and thus security of the code. We bring it back (from old snapshots?) when needed.

contracts/rosca.sol
rmerom: right, if we leave it in, we'll use a parameter when testing. However, given the high stakes of what we're trying to do (an immutable contract that holds real money), we might want to consider starting with the simplest case (e.g. 1 epoch per ROSCA), getting to the highest confidence that we got it right, then adding the extra features. There's just no way going back if we screw up...

In fact, let's speak with George, I'm not sure if we need >1 epoch for MVP.

contracts/rosca.sol
rmerom: oh sorry, I missed that. I think we should fix the comment then.

contracts/rosca.sol
rmerom: You're absolutely right, i missed that.
Even then, let's consider not throwing when someone is doing something non-maliciously - we're just passing money (gas) from them to miners.

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: ok

contracts/rosca.sol
shine2lay: fixed

contracts/rosca.sol
shine2lay: fixed

contracts/rosca.sol
shine2lay: done

contracts/ROSCA.sol
shine2lay: one is put there by you, during our meeting at starbucks , its for triggering escape hatch
should we take it out completely?

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: done

contracts/rosca.sol
shine2lay: i understand what you mean by deleting all the remaining request, i dont get the part about gas refund to the caller

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
rmerom: Ah, got it . How about we remove it for now, as it's a dead code?

contracts/rosca.sol
rmerom: i've sent an email to product-discuss with this question, please voice your opinion if you're thinking we're going in a wrong direction.

contracts/ROSCA.sol

contracts/rosca.sol
rmerom: when we a call clears a storage variable, gas is refunded to the caller (see under the heading "gas refund" here: https://media.consensys.net/2016/06/23/ethereum-gas-fuel-and-fees/ )

contracts/rosca.sol
rmerom: Please add a short comment explaining what this contract is.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+/**
+*  Things still missing/needs fixing:
+*   in withdraw() , add fee related logic
+*   checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+*/
+contract ROSCA {
+  address constant WETRUST = 0x0;
+  uint constant MIN_CONTRIBUTION_SIZE = 1000000000000;  // 1e12
+  uint constant MAX_FEE_IN_THOUSANDTHS = 200;
+  address constant WETRUST_FEE_ADDRESS = 0x0;
```
rmerom: Let's a TODO here to update the address

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+  uint constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;
```
rmerom: consider making this uint32, so if we use it in a smaller state variable, we won't need to convert. Same about considering choosing smaller sizes for the other constants.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+  uint constant MINIMUM_PARTICIPANTS = 2;
+  uint16 constant EPOCH_TO_RUN = 1;
+  uint constant MIN_ROUND_PERIOD_IN_DAYS = 1;
+  uint constant MAX_ROUND_PERIOD_IN_DAYS = 30;
+  uint constant MIN_DISTRIBUTION_RATIO = 65;
```
rmerom: Please add an explanation what this is.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+*   checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+*/
+contract ROSCA {
+  address constant WETRUST = 0x0;
+  uint constant MIN_CONTRIBUTION_SIZE = 1000000000000;  // 1e12
+  uint constant MAX_FEE_IN_THOUSANDTHS = 200;
+  address constant WETRUST_FEE_ADDRESS = 0x0;
+  uint constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;
+  uint constant MINIMUM_PARTICIPANTS = 2;
+  uint16 constant EPOCH_TO_RUN = 1;
```
rmerom: Let's talk about this

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+  event LogStartOfPeriod(uint currentRound);
+
+  // state variables
+  uint16 currentEpoch;
+  uint16 numEpochs;     // number of Epochs this contract will run
+  uint16 roundPeriodInDays;
+  uint16 serviceFeeInThousandths;
+  uint16 currentRound;  // currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually starts
+  uint8 minParticipants;
+  address foreman;    // address of the foreman
```
rmerom: comment may be redundant

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+  uint contributionSize;
+  uint startTime;     // startTime must be in timestamp
```
rmerom: comment may be redundant

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+  struct User{
```
rmerom: space before {

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    uint contributed; // Total amount contributed
+    uint paid; // number of times bid Won
+    uint pendingWithdrawl; // how much they are allowed to withdraw, i.e if someone wins , their pendingWithdrawal will go up by the bid.
+    bool alive; // needed to check if a member is indeed a member
+  }
+  mapping(address => User) members; // using struct User to keep track of contributions and paid, allowance and etc.
```
rmerom: Please add blank line after end of struct

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+  mapping(address => bool) pendingJoinRequest; // this way , address can be used as index, if we use address[] , we'll have to go thru a whole array
+
+  // bidding related state variable
+  uint lowestBid;
+  address winnerAddress;
+
+
+  modifier onlyForeman {
+  if(msg.sender != foreman) throw;
```
rmerom: please indent code in modifier.
same in modifier below.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+  function ROSCA(
+    uint16 roundPeriodInDays_,
+    uint contributionSize_,
+    uint8 minParticipants_,
+    uint startTime_,
+    uint16 feeInThousandths_)
+  {
+    if (roundPeriodInDays < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodInDays > MAX_ROUND_PERIOD_IN_DAYS) throw;
+    roundPeriodInDays = roundPeriodInDays_;
+    if (contributionSize_ < MIN_CONTRIBUTION_SIZE) throw;
```
rmerom: just thinking aloud: maybe we should also have a MAX_CONTRIBUTION_SIZE because we don't want to hold thousands of ethers right now in an MVP (due to security restrictions)? what do you think?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    contributionSize = contributionSize_;
+
+    numEpochs = EPOCH_TO_RUN;
```
rmerom: if we're not using epochs right now, i suggest we remove any reference to them. we wanted to remove them because of code complexity which might affect security, leaving them in there will not get us that benefit.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+    if (minParticipants_ < MINIMUM_PARTICIPANTS) throw;// there should be at least 2 people to make a group
```
rmerom: space before //

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    minParticipants = minParticipants_;
+
+    if (startTime_ < (now + MINIMUM_TIME_BEFORE_ROSCA_START)) throw;
+    startTime = startTime_;
+
+    if (feeInThousandths_ > MAX_FEE_IN_THOUSANDTHS) throw;
+    serviceFeeInThousandths = feeInThousandths_;
```
rmerom: suggestion: call feeInThousands_ serviceFeeInThousandths_ (the same as the state variable), or otherwise remove the _ (as there is no name conflict).

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+
+    // set foreman
+    foreman = msg.sender;
+
+    // register foreman as a member of ROSCA and increment numRounds in a Epoch to 1
```
rmerom: I think the second part of the comment is no longer relevant.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    members[msg.sender] = User({paid: 0 , contributed: 0, alive: true, pendingWithdrawl: 0});
```
rmerom: suggestion: extract these 2 lines into an internal method 
addMember(address member)
to avoid code duplication (with the code when approving members)

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    membersAddresses.push(msg.sender);
+  }
+
+  /** startRound() check if the round has ended, if so, get the winner of the currentRound's pot
+    * and add the amount to the winner's pending withdrawal
+    * if there were no bid during the round, address at selected at semi-random
```
rmerom: remove "at"

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    *
+    * if currentRound = 0(ROSCA hasnt started), no winner is declared
+    */
+
+  function startRound()
+  {
+
+    if (now < startTime + (uint(currentRound + (currentEpoch * membersAddresses.length))  * (uint(roundPeriodInDays) * 1 days)) || membersAddresses.length < minParticipants)
```
rmerom: Consider extracting the calculation into a local var:

uint roundStartTime = startTime + (uint(....  ;

then

if (now < roundStartTime || membersAddresses.length < minParticipants) ...

so that it's easier to read and spot any potential bugs.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+  /** startRound() check if the round has ended, if so, get the winner of the currentRound's pot
+    * and add the amount to the winner's pending withdrawal
+    * if there were no bid during the round, address at selected at semi-random
+    *
+    * if currentRound = 0(ROSCA hasnt started), no winner is declared
+    */
+
+  function startRound()
+  {
```
rmerom: { 
should be on previous line

same below in some of the if's/for's

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+
+    if (now < startTime + (uint(currentRound + (currentEpoch * membersAddresses.length))  * (uint(roundPeriodInDays) * 1 days)) || membersAddresses.length < minParticipants)
+      throw;
+
+    if (currentRound != 0) {
+      if (winnerAddress == 0) //only true when there is no bidder in this round
```
rmerom: space after //

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+      {
+        // there is no bid in this round so find an unpaid address for this epoch
+        for (uint i = 0 ; i < membersAddresses.length; i++)
```
rmerom: extra space before ;

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+        {
+          uint semi_random = now % membersAddresses.length;
```
rmerom: this var isn't used. it should probably be initialized before the for, then incremented in every iteration.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+        }
+      }
+      if(members[winnerAddress].paid == currentEpoch) throw;
+      members[winnerAddress].pendingWithdrawl += lowestBid;
+      members[winnerAddress].paid++;
+      LogRoundFundsReleased(winnerAddress, lowestBid);
+    }
+    if (currentRound < membersAddresses.length)  // reset variables related to bidding
+    {
+      lowestBid = contributionSize * membersAddresses.length;
```
rmerom: you might want to add +1 to it, so that if just one person wants the money, then can get it without paying any fee.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+      winnerAddress = 0;
+
+      currentRound++;
+    }else if (currentEpoch < numEpochs) // end of Epoch reset rounds
```
rmerom: space after }

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+      currentEpoch++;
+      currentRound = 1;
+      lowestBid = contributionSize;
+    }
+  }
+
+  /**
+   * Anyone not already a member of ROSCA can request to join and they'll be put into
+   * pendingJoinRequest untill foreman accept request or ROSCA has started
+   *
```
rmerom: extra line

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    {
+      currentEpoch++;
+      currentRound = 1;
+      lowestBid = contributionSize;
+    }
+  }
+
+  /**
+   * Anyone not already a member of ROSCA can request to join and they'll be put into
+   * pendingJoinRequest untill foreman accept request or ROSCA has started
```
rmerom: s/untill/until
s/accept, request/accepts, requests

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+   *
+   */
+  function joinRequest() onlyBeforeStart
+  {
+    // only put the request in the pending list if they are not in the ROSCA already
+    if (members[msg.sender].alive) throw;
```
rmerom, line 144: giving it another thought: do we really want to throw here? if they've been accepted to the ROSCA, they're not doing anything malicious by sending a request to join (e.g. maybe there's a bug in our app or our backend).
Shouldn't we just return instead of throw?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    pendingJoinRequest[msg.sender] = true;
+    LogParticipantApplied(msg.sender);
+  }
+
+  /**
+   *
```
rmerom: Extra line

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+   * once a requestor had been registered as member, the address will be taken out of pendingJoinRequest
```
rmerom, line 151: This is one detail about how the function works.
Perhaps we should document it for users, e.g. "foreman can call this method to approve a join request by a participant"?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+   **/
+  function acceptJoinRequest(address requestor)
+    onlyForeman
+    beforeStart
+  {
+    if (!pendingJoinRequest[requestor]) throw;
```
rmerom, line 156: same argument as above for returning() instead of throwing

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    members[requestor] = USER({paid: 0 , contributed: 0, alive: true, pendingWithdrawl: 0});
+    membersAddresses.push(requestor);
+    numRoundsInEpoch++;
```
rmerom: Does this compile? i thought the var was commented out.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    delete(pendingJoinRequest[requestor]); // take out the requestor's address in the pending list
+  }
+
+  /**
+   * Processes a periodic contribution from msg.sender ().
+   * Any excess funds will be withdrawable through withdraw().
+   */
+  function contribute() payable
+  {
+    if (members[msg.sender].alive)
```
rmerom: we need to throw if (!members[msg.sender].alive), otherwise there'll be funds stuck in the contract

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  /**
+   * Registers a bid from msg.sender. If msg.sender has already won a round or bid is higher than lowestBid,
+   * this method will throw.
+   */
+  function bid(uint distrubtionAmountInWei)
+  {
+    if (!(distrubtionAmountInWei < lowestBid && members[msg.sender].paid < currentEpoch && currentRound > 0
```
rmerom: - Instead of if (!(a > x && b < y && c > z)) consider using if (a <= x || b >= y || c <= z). It might be easier to read.
  One way or another, this is pretty complex logic, and our penalty for mistakes is huge (users losing money, we lost reputation). Consider breaking this into a few simpler lines.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    && distrubtionAmountInWei < (((contributionSize * membersAddresses.length)/100) * MIN_DISTRIBUTION_RATIO))) throw;
+    lowestBid = distrubtionAmountInWei;
+    winnerAddress = msg.sender;
+    LogNewLowestBid(lowestBid, winnerAddress);
+  }
+
+  /**
+   * Withdraws available funds for msg.sender. If opt_destination is specified,
```
rmerom, line 187: s/for/to

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+   * sends the fund to that address.
+   */
+  function withdraw(address opt_destination)
+    returns(bool success)
+  {
+    if (opt_destination == 0)
+      opt_destination = msg.sender;
+    if (!members[msg.sender].alive || members[msg.sender].pendingWithdrawl > 0) throw;
```
rmerom: s/msg.sender/opt_destination

same in the rest of this function

rmerom: why do we throw if pendingWithdrawl > 0 ? shouldn't it be the other way around?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,222 @@
...
+    uint amountToWithdraw = members[msg.sender].pendingWithdrawl;  // use a temporary variable to avoid the receiver calling the withdraw function again before
+    members[msg.sender].pendingWithdrawl = 0;                      // send() completes its task
+    if (!opt_destination.send(amountToWithdraw)) {   // if the send() fails, put the allowance back to its original place
+      // No need to call throw here, just reset the amount owing
+     members[msg.sender].pendingWithdrawl = amountToWithdraw;
```
rmerom: fix indentation

rmerom (commented):
The code looks really amazing now! just a few more things

contracts/rosca.sol
shine2lay: done

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: i think taking that out now would take more work

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: sure, i don't see why not

shine2lay: done, set it to 10 eth upper limit for now, we can change it later

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: true,done

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: i think throw is better than returning in most cases, we can discuss more if you want

contracts/ROSCA.sol
shine2lay: taken out

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: umm, msg.sender is right 
opt_destination isnt necessarily a member

yes, pendingwithdrawl == 0 should be the check

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: for is right, available funds for that member, to opt_destination if specified

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: done

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: i used the constant EPOCH_TO_RUN as check and took out any reference to numEpochs or currentEpoch

contracts/ROSCA.sol
rmerom: you're right

contracts/ROSCA.sol
rmerom: right, msg.sender it is!

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+ * Things still missing/needs fixing:
+ * in withdraw() , add fee related logic
+ * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+ */
+contract ROSCA {
+  uint32 constant MIN_CONTRIBUTION_SIZE = uint32(1000000000000);  // 1e12
+  uint32 constant MAX_CONTRIBUTION_SIZE = uint32(10000000000000000000); // 10 ether in Wei
+  uint32 constant MAX_FEE_IN_THOUSANDTHS = 200;
+  address constant WETRUST_FEE_ADDRESS = 0x0;           // TODO: needs to be updated
+  uint32 constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;   // startTime of the ROSCA must be at least 1 day away from now
```
rmerom: consider 
s/now/from when the ROSCA is created

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+  uint32 constant MINIMUM_PARTICIPANTS = 2;           // minimum participants for the ROSCA to start
+  uint16 constant EPOCH_TO_RUN = 1;
```
rmerom: this might be hard to understand for someone who doesn't know the history of the code.
Consider changing "paid" field of User to be boolean and removing this field.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+  uint32 constant MAX_ROUND_PERIOD_IN_DAYS = 30;
+  uint32 constant MIN_DISTRIBUTION_RATIO = 65;  // the winning bid must be at least 65% of the Pot value
+
+  event LogParticipantApplied(address user);
+  event LogParticipantApproved(address user);
+  event LogContributionMade(address user, uint amount);
+  event LogNewLowestBid(uint bid,address winnerAddress);
+  event LogRoundFundsReleased(address winnerAddress, uint amountInWei);
+  event LogFundsWithdrawal(address user, uint amount,address destination);
+  event LogStartOfPeriod(uint currentRound);
```
rmerom: This isn't used. Consider renaming to LogStartOfRound and adding to startRound()

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+
+  // state variables
+  uint16 roundPeriodInDays;
+  uint16 serviceFeeInThousandths;
+  uint16 currentRound;  // currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually starts
+  uint8 minParticipants;
```
rmerom: Hmmm I'm cringing a bit about this being uint8 (seems like a pitfall in case someone passes a number > 255). Can we change this to uint16? it won't take more space because all of these small ints will still be bundled into one 256 bit word.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+    * Creates a new ROSCA and initializes the necessary variables, ROSCA doesnt start until the specified startTime
+    * Creator of the contract becomes foreman and also added as the first member of the ROSCA
+    */
+  function ROSCA(
+    uint16 roundPeriodInDays_,
+    uint contributionSize_,
+    uint8 minParticipants_,
+    uint startTime_,
+    uint16 serviceFeeInThousandths_)
+  {
```
rmerom: Please move brace to previous line, same in some of the functions below

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+    minParticipants = minParticipants_;
+
+    if (startTime_ < (now + MINIMUM_TIME_BEFORE_ROSCA_START)) throw;
+    startTime = startTime_;
+
+    if (serviceFeeInThousandths_ > MAX_FEE_IN_THOUSANDTHS) throw;
+    serviceFeeInThousandths = serviceFeeInThousandths_;
+
+
+    // set foreman
```
rmerom: redundant comment? (code is self-explanatory)
same in L97.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+      if (winnerAddress == 0) { // only true when there is no bidder in this round
+        // there is no bid in this round so find an unpaid address for this epoch
+        uint semi_random = now % membersAddresses.length;
+        for (uint i = 0; i < membersAddresses.length; i++) {
+          if(members[membersAddresses[(semi_random + i) % membersAddresses.length]].paid < EPOCH_TO_RUN)
+            winnerAddress = membersAddresses[semi_random + i];
+          break;
+        }
+      }
+      if(members[winnerAddress].paid == EPOCH_TO_RUN) throw;
```
rmerom: shouldn't these lines be inside the "if (currentRound != 0)" ?
Actually, if currentRound == 0, can't we just currentRound++ and exit the function?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  /**
+   * Anyone not already a member of ROSCA can request to join and they'll be put into
+   * pendingJoinRequest until foreman accepts requests or ROSCA has started
+   */
+  function joinRequest()
+    onlyBeforeStart
+  {
```
rmerom: opening brace on same line, same in functions below. Might want to put the entire function header in a single line, if it fits.
Same for functions below.

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+    members[msg.sender].contributed += msg.value;
+    LogContributionMade(msg.sender, msg.value);
+  }
+
+  /**
+   * Registers a bid from msg.sender. If msg.sender has already won a round or bid is higher than lowestBid,
+   * this method will throw.
+   */
+  function bid(uint distrubtionAmountInWei) {
+    if (distrubtionAmountInWei >= lowestBid || members[msg.sender].paid >= EPOCH_TO_RUN || currentRound == 0
```
rmerom: - I think we want to prevent the user from asking a distribution that is < the MIN_DISTRIBUTION_RATIO, not >
- Consider breaking this into a few simpler lines. Specifically, it'd be nice to break the complex calculation into another line.
- Lot of codebases use 4-line indentation for long lines, to it's easy to see that a line was broken, e.g.
  if (someVeryLongCondition && anotherCondition
  && yetAnotherCondition) {
  doSomething;
  }

Would you be ok doing that?

contracts/ROSCA.sol
```diff
@@ -0,0 +1,218 @@
...
+      opt_destination = msg.sender;
+    if (!members[msg.sender].alive || members[msg.sender].pendingWithdrawl == 0) throw;
+    uint amountToWithdraw = members[msg.sender].pendingWithdrawl;  // use a temporary variable to avoid the receiver calling the withdraw function again before
+    members[msg.sender].pendingWithdrawl = 0;                      // send() completes its task
+    if (!opt_destination.send(amountToWithdraw)) {   // if the send() fails, put the allowance back to its original place
+      // No need to call throw here, just reset the amount owing
+      members[msg.sender].pendingWithdrawl = amountToWithdraw;
+      return false;
+    }
+    else{
```
rmerom: - else on same line as }
- space before {

migrations/2_deploy_contracts.js
```diff
@@ -1,5 +1,5 @@
 module.exports = function(deployer) {
   deployer.deploy(ConvertLib);
```
rmerom: do we need ConvertLib ?

rmerom (approved):
Great work Shine!
Some nitpicks, and 1-2 more important comments, please take a look.

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: remove

contracts/ROSCA.sol

contracts/ROSCA.sol

migrations/2_deploy_contracts.js
shine2lay: idk what that is for

contracts/ROSCA.sol
shine2lay: i dont think we need it , members only allowed to place bet if they havent been paid

contracts/ROSCA.sol
shine2lay: > = is right , this is reverse auction, if they bid higher than lowest bid, it should throw

migrations/2_deploy_contracts.js
rmerom: It's an example contract  that is added when calling "truffle init". It's located under contracts and I don't think we need it.

rmerom (approved):

PR #7, opened 2016-11-23
latest ROSCA.sol

shine2lay:
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/7)

### DIFF
--- a/contracts/ROSCA.sol
+++ b/contracts/ROSCA.sol
@@ -1,267 +1,259 @@
- pragma solidity ^0.4.4;
+pragma solidity ^0.4.4;
 
- /**
-  * ROSCA contract allows participants to get into an agreement with each other to contribute a certain amount of ether every round
-  * In every round, one participant will recieve ether that everyone contributed.
-  * The winner of the pot is decided by reverse auction (lowest Bid wins).
-  *
-  * Things still missing/needs fixing:
-  * in withdraw() , add fee related logic (done, Sol should still be discussed)
-  *   Sol: will keep deduct a fee, and keep it in contract untill the end?
-  * if there are contribution than what is required, put them in pendingWithdrawal (done)
-  *   Sol: in contribute , check if contributed is enough to last the whole ROSCA, if it is put the excess in pendingWithdrawal
-  * when the rosca ends all the money should go out to the account user (done, still needs to work out sending fees to wetrust address)
-  *   Sol: added endOfROSCA, cleanUp and sendFUND function
-  * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round (goodStanding var takes care of this i think)
-  *  -startRound should check?
-  * if user doesnt contribute , dont allow bidding EVER! have to be in good standing to bid/win the Pot (done)
-  *   Sol: used goodStanding variable to keep track of how many round a person can bet
-  * contribute and withdraw doesnt check if round has started (done)
-  *
-  * Product Decision to think about:
-  *   - biding when not everyone had contributed, total contributed is less than the bid amount
-  *     Possible Solution:
-  *       - owed variable? before paying the next round winner, owed is paid first?
-  *       - startRound Check ? if balance is less than lowest Bid ? if it is, what to do?
-  *       - in withdraw, check balance of contract first and if its less than pending withdrawal, subtract that amount only (I think this is the simplest option) (implement for now)
-  *
-  * Things brought up that needs considering:
-  *   - allowing constructor to have initial list of members specified by foreman on ROSCA deployment
-  *   - using a better unit than wei to represent ether i.e finney (we shouldn't be dealing with any value less than finney anyways)
-  *   - getters for necessary variables (making variables public would work)
-  *   - optimizing , contract deployment currently cost over 20 cents on test-net
-  *   - someway to kill the contract when ROSCA is ended? (currently cleanUp functions includes selfdestruct)
-  *   - frontend should try a call function and if it throws, warns user of possible throw before allowing transaction.
-  */
- contract ROSCA {
-   uint64 constant MIN_CONTRIBUTION_SIZE = 1 finney;  // 1e12
-   uint64 constant MAX_CONTRIBUTION_SIZE = 10 ether;//uint64(10000000000000000000); // 10 ether in Wei
-   uint16 constant MAX_FEE_IN_THOUSANDTHS = 200;
-   uint32 constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;   // startTime of the ROSCA must be at least 1 day away from when the ROSCA is created
-   uint8 constant MINIMUM_PARTICIPANTS = 2;           // minimum participants for the ROSCA to start
-   uint8 constant MIN_ROUND_PERIOD_IN_DAYS = 1;
-   uint8 constant MAX_ROUND_PERIOD_IN_DAYS = 30;
-   uint8 constant MIN_DISTRIBUTION_RATIO = 65;  // the winning bid must be at least 65% of the Pot value
-   address constant WETRUST_FEE_ADDRESS = 0x0;           // TODO: needs to be updated
+/**
+ * ROSCA contract allows participants to get into an agreement with each other to contribute a certain amount of ether every round
+ * In every round, one participant will recieve ether that everyone contributed.
+ * The winner of the pot is decided by reverse auction (lowest Bid wins).
+ *
+ * Things still missing/needs fixing:
+ * in withdraw() , add fee related logic (done, Sol should still be discussed)
+ *   Sol: will keep deduct a fee, and keep it in contract untill the end?
+ * if there are contribution than what is required, put them in pendingWithdrawal (done)
+ *   Sol: in contribute , check if contributed is enough to last the whole ROSCA, if it is put the excess in pendingWithdrawal
+ * when the rosca ends all the money should go out to the account user (done, still needs to work out sending fees to wetrust address)
+ *   Sol: added endOfROSCA, cleanUp and sendFUND function
+ * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round (goodStanding var takes care of this i think)
+ *  -startRound should check?
+ * if user doesnt contribute , dont allow bidding EVER! have to be in good standing to bid/win the Pot (done)
+ *   Sol: used goodStanding variable to keep track of how many round a person can bet
+ * contribute and withdraw doesnt check if round has started (done)
+ *
+ * Product Decision to think about:
+ *   - biding when not everyone had contributed, total contributed is less than the bid amount
+ *     Possible Solution:
+ *       - owed variable? before paying the next round winner, owed is paid first?
+ *       - startRound Check ? if balance is less than lowest Bid ? if it is, what to do?
+ *       - in withdraw, check balance of contract first and if its less than pending withdrawal, subtract that amount only (I think this is the simplest option) (implement for now)
+ *
+ * Things brought up that needs considering:
+ *   - allowing constructor to have initial list of members specified by foreman on ROSCA deployment
+ *   - using a better unit than wei to represent ether i.e finney (we shouldn't be dealing with any value less than finney anyways)
+ *   - getters for necessary variables (making variables public would work)
+ *   - optimizing , contract deployment currently cost over 20 cents on test-net
+ *   - someway to kill the contract when ROSCA is ended? (currently cleanUp functions includes selfdestruct)
+ *   - frontend should try a call function and if it throws, warns user of possible throw before allowing transaction.
+ */
+contract ROSCA {
+  uint64 constant MIN_CONTRIBUTION_SIZE = 1 finney;  // 1e12
+  uint64 constant MAX_CONTRIBUTION_SIZE = 10 ether;//uint64(10000000000000000000); // 10 ether in Wei
+  uint16 constant MAX_FEE_IN_THOUSANDTHS = 200;
+  uint32 constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;   // startTime of the ROSCA must be at least 1 day away from when the ROSCA is created
+  uint8 constant MINIMUM_PARTICIPANTS = 2;           // minimum participants for the ROSCA to start
+  uint8 constant MIN_ROUND_PERIOD_IN_DAYS = 1;
+  uint8 constant MAX_ROUND_PERIOD_IN_DAYS = 30;
+  uint8 constant MIN_DISTRIBUTION_RATIO = 65;  // the winning bid must be at least 65% of the Pot value
+  address constant WETRUST_FEE_ADDRESS = 0x0;           // TODO: needs to be updated
 
+  event LogParticipantApplied(address user);
+  event LogParticipantApproved(address user);
+  event LogContributionMade(address user, uint amount);
+  event LogNewLowestBid(uint bid,address winnerAddress);
+  event LogRoundFundsReleased(address winnerAddress, uint amountInWei);
+  event LogFundsWithdrawal(address user, uint amount,address destination);
+  event LogStartOfRound(uint currentRound);
 
-   event LogParticipantApplied(address user);
-   event LogParticipantApproved(address user);
-   event LogContributionMade(address user, uint amount);
-   event LogNewLowestBid(uint bid,address winnerAddress);
-   event LogRoundFundsReleased(address winnerAddress, uint amountInWei);
-   event LogFundsWithdrawal(address user, uint amount,address destination);
-   event LogStartOfRound(uint currentRound);
+  // state variables
+  uint16 roundPeriodInDays;
+  uint16 serviceFeeInThousandths;
+  uint16 currentRound;  // currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually starts
+  uint16 minParticipants;
+  bool endOfROSCA = false;
+  address foreman;
+  uint32 contributionSize;
+  uint startTime;
 
-   // state variables
-   uint16 roundPeriodInDays;
-   uint16 serviceFeeInThousandths;
-   uint16 currentRound;  // currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually starts
-   uint16 minParticipants;
-   bool endOfROSCA = false;
-   address foreman;
-   uint32 contributionSize;
-   uint startTime;
+  struct User {
+    uint contributed; // Total amount contributed
+    bool paid; // yes if the member had been paid
+    uint pendingWithdrawal; // how much they are allowed to withdraw, i.e if someone wins , their pendingWithdrawal will go up by the bid.
+    bool alive; // needed to check if a member is indeed a member
+    uint goodStanding;
+  }
 
-   struct User {
-     uint contributed; // Total amount contributed
-     bool paid; // yes if the member had been paid
-     uint pendingWithdrawal; // how much they are allowed to withdraw, i.e if someone wins , their pendingWithdrawal will go up by the bid.
-     bool alive; // needed to check if a member is indeed a member
-     uint goodStanding;
-   }
+  mapping(address => User) members; // using struct User to keep track of contributions and paid, allowance and etc.
+  address[] membersAddresses;    // this is the only way to iterate through all the member's address
 
-   mapping(address => User) members; // using struct User to keep track of contributions and paid, allowance and etc.
-   address[] membersAddresses;    // this is the only way to iterate through all the member's address
-
-   mapping(address => bool) pendingJoinRequest; // this way , address can be used as index, if we use address[] , we'll have to go thru a whole array
-
+  mapping(address => bool) pendingJoinRequest; // this way , address can be used as index, if we use address[] , we'll have to go thru a whole array
    // bidding related state variable
-   uint lowestBid;
-   address winnerAddress;
+  uint lowestBid;
+  address winnerAddress;
 
+  modifier onlyForeman {
+    if(msg.sender != foreman) throw;
+    _;
+  }
+  modifier onlyBeforeStart {
+    if(currentRound != 0) throw;
+    _;
+  }
 
-   modifier onlyForeman {
-     if(msg.sender != foreman) throw;
-     _;
-   }
-   modifier onlyBeforeStart {
-     if(currentRound != 0) throw;
-     _;
-   }
+  /**
+    * Creates a new ROSCA and initializes the necessary variables, ROSCA doesnt start until the specified startTime
+    * Creator of the contract becomes foreman and also added as the first member of the ROSCA
+    */
+  function ROSCA(
+    uint16 roundPeriodInDays_,
+    uint32 contributionSize_,
+    uint16 minParticipants_,
+    uint startTime_,
+    uint16 serviceFeeInThousandths_) {
+    if (roundPeriodInDays_ < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodInDays_ > MAX_ROUND_PERIOD_IN_DAYS) throw;
+    roundPeriodInDays = roundPeriodInDays_;
+    if (contributionSize_ < MIN_CONTRIBUTION_SIZE || contributionSize_ > MAX_CONTRIBUTION_SIZE) throw;
+    contributionSize = contributionSize_;
 
-   /**
-     * Creates a new ROSCA and initializes the necessary variables, ROSCA doesnt start until the specified startTime
-     * Creator of the contract becomes foreman and also added as the first member of the ROSCA
-     */
-   function ROSCA(
-     uint16 roundPeriodInDays_,
-     uint32 contributionSize_,
-     uint16 minParticipants_,
-     uint startTime_,
-     uint16 serviceFeeInThousandths_) {
-     if (roundPeriodInDays_ < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodInDays_ > MAX_ROUND_PERIOD_IN_DAYS) throw;
-     roundPeriodInDays = roundPeriodInDays_;
-     if (contributionSize_ < MIN_CONTRIBUTION_SIZE || contributionSize_ > MAX_CONTRIBUTION_SIZE) throw;
-     contributionSize = contributionSize_;
-
-
-     if (minParticipants_ < MINIMUM_PARTICIPANTS) throw; // there should be at least 2 people to make a group
-     minParticipants = minParticipants_;
-
+    if (minParticipants_ < MINIMUM_PARTICIPANTS) throw; // there should be at least 2 people to make a group
+    minParticipants = minParticipants_;
      if (startTime_ < (now + MINIMUM_TIME_BEFORE_ROSCA_START)) throw;
-     startTime = startTime_;
-
+    startTime = startTime_;
      if (serviceFeeInThousandths_ > MAX_FEE_IN_THOUSANDTHS) throw;
-     serviceFeeInThousandths = serviceFeeInThousandths_;
+    serviceFeeInThousandths = serviceFeeInThousandths_;
 
+    foreman = msg.sender;
+    addMember(msg.sender);
+  }
 
-     foreman = msg.sender;
-     addMember(msg.sender);
-   }
+  function addMember(address newMember) internal {
+    members[newMember] = User({paid: false , contributed: 0, alive: true, pendingWithdrawal: 0,goodStanding: 0});
+    membersAddresses.push(newMember);
+  }
 
-   function addMember(address newMember) internal {
-     members[newMember] = User({paid: false , contributed: 0, alive: true, pendingWithdrawal: 0,goodStanding: 0});
-     membersAddresses.push(newMember);
-   }
-
-   /** startRound() check if the round has ended, if so, get the winner of the currentRound's pot
-     * and add the amount to the winner's pending withdrawal
-     * if there were no bid during the round, address selected at semi-random
-     *
-     * if currentRound = 0(ROSCA hasnt started), no winner is declared
-     */
-
-   function startRound() {
-     uint roundStartTime = startTime + (uint(currentRound)  * (uint(roundPeriodInDays) * 1 days));
-     if (now < roundStartTime || membersAddresses.length < minParticipants)
-       throw;
-
-     if (currentRound != 0) {
-       if (winnerAddress == 0) { // only true when there is no bidder in this round
-         // there is no bid in this round so find an unpaid address for this epoch
-         uint semi_random = now % membersAddresses.length;
-         for (uint i = 0; i < membersAddresses.length; i++) {
-           if(!members[membersAddresses[(semi_random + i) % membersAddresses.length]].paid)
-             winnerAddress = membersAddresses[semi_random + i];
-           break;
-         }
-       }
-       members[winnerAddress].pendingWithdrawal += (lowestBid / 100000) * MAX_FEE_IN_THOUSANDTHS;
-       members[winnerAddress].paid = true;
-       LogRoundFundsReleased(winnerAddress, lowestBid);
-     }
-     if (currentRound < membersAddresses.length) {  // reset variables related to bidding
-       lowestBid = contributionSize * membersAddresses.length + 1;
-       winnerAddress = 0;
-
-       currentRound++;
-       LogStartOfRound(currentRound);
-     } else if (currentRound == membersAddresses.length) {
-         endOfROSCA = true;
-     }
-   }
-
-   /**
-    * try to send the fund to the opt_destination from source account
-    */
-   function sendFund(address source,address opt_destination) internal returns(bool success){
-     uint amountToWithdraw = members[source].pendingWithdrawal  ;  // use a temporary variable to avoid the receiver calling the withdraw function again before
-     if(this.balance < amountToWithdraw) amountToWithdraw = this.balance;
-     members[source].pendingWithdrawal = members[source].pendingWithdrawal - amountToWithdraw;                      // send() complete its task
-     if (!opt_destination.send(amountToWithdraw)) {   // if the send() fails, put the allowance back to its original place
-       // No need to call throw here, just reset the amount owing
-       members[source].pendingWithdrawal = amountToWithdraw;
-       return false;
-     } else {
-       LogFundsWithdrawal(source, amountToWithdraw, opt_destination);
-       return true;
-     }
-   }
-
-   /**
-    * cleanUp is to make sure that fees are send to where they are owed and
-    * make sure contract doesn't hold any ether after it ends.
+  /** startRound() check if the round has ended, if so, get the winner of the currentRound's pot
+    * and add the amount to the winner's pending withdrawal
+    * if there were no bid during the round, address selected at semi-random
     *
+    * if currentRound = 0(ROSCA hasnt started), no winner is declared
     */
-   function cleanUp() {
-     if(!endOfROSCA || this.balance == 0) throw;
+
+  function startRound() {
+    uint roundStartTime = startTime + (uint(currentRound)  * (uint(roundPeriodInDays) * 1 days));
+    if (now < roundStartTime || membersAddresses.length < minParticipants)
+      throw;
+
+    if (currentRound != 0) {
+      if (winnerAddress == 0) { // only true when there is no bidder in this round
+        // there is no bid in this round so find an unpaid address for this epoch
+        uint semi_random = now % membersAddresses.length;
+        for (uint i = 0; i < membersAddresses.length; i++) {
+          if(!members[membersAddresses[(semi_random + i) % membersAddresses.length]].paid)
+            winnerAddress = membersAddresses[semi_random + i];
+          break;
+        }
+      }
+      members[winnerAddress].pendingWithdrawal += (lowestBid / 100000) * MAX_FEE_IN_THOUSANDTHS;
+      members[winnerAddress].paid = true;
+      LogRoundFundsReleased(winnerAddress, lowestBid);
+    }
+    if (currentRound < membersAddresses.length) {  // reset variables related to bidding
+      lowestBid = contributionSize * membersAddresses.length + 1;
+      winnerAddress = 0;
+
+      currentRound++;
+      LogStartOfRound(currentRound);
+    } else if (currentRound == membersAddresses.length) {
+        endOfROSCA = true;
+    }
+  }
+
+  /**
+   * try to send the fund to the opt_destination from source account
+   */
+  function sendFund(address source,address opt_destination) internal returns(bool success){
+    uint amountToWithdraw = members[source].pendingWithdrawal  ;  // use a temporary variable to avoid the receiver calling the withdraw function again before
+    if(this.balance < amountToWithdraw) amountToWithdraw = this.balance;
+    members[source].pendingWithdrawal = members[source].pendingWithdrawal - amountToWithdraw;                      // send() complete its task
+    if (!opt_destination.send(amountToWithdraw)) {   // if the send() fails, put the allowance back to its original place
+      // No need to call throw here, just reset the amount owing
+      members[source].pendingWithdrawal = amountToWithdraw;
+      return false;
+    } else {
+      LogFundsWithdrawal(source, amountToWithdraw, opt_destination);
+      return true;
+    }
+  }
+
+  /**
+   * cleanUp is to make sure that fees are send to where they are owed and
+   * make sure contract doesn't hold any ether after it ends.
+   *
+   */
+  function cleanUp() {
+    if(!endOfROSCA || this.balance == 0) throw;
 
      for(uint8 i = 0; i < membersAddresses.length; i++) {
-       if(members[membersAddresses[i]].pendingWithdrawal > 0){
-         sendFund(membersAddresses[i],membersAddresses[i]);
-       }
-     }
-     //send the rest to WETRUST_FEE_ADDRESS because all that is left should be the fees to service
-     if(!WETRUST_FEE_ADDRESS.send(this.balance))
-         throw;
-   }
+      if(members[membersAddresses[i]].pendingWithdrawal > 0){
+        sendFund(membersAddresses[i],membersAddresses[i]);
+      }
+    }
+    //send the rest to WETRUST_FEE_ADDRESS because all that is left should be the fees to service
+    if(!WETRUST_FEE_ADDRESS.send(this.balance))
+        throw;
+  }
 
-   /**
-    * Anyone not already a member of ROSCA can request to join and they'll be put into
-    * pendingJoinRequest until foreman accepts requests or ROSCA has started
-    */
-   function joinRequest() onlyBeforeStart {
-     // only put the request in the pending list if they are not in the ROSCA already
-     if (members[msg.sender].alive) throw;
-     pendingJoinRequest[msg.sender] = true;
-     LogParticipantApplied(msg.sender);
-   }
+  /**
+   * Anyone not already a member of ROSCA can request to join and they'll be put into
+   * pendingJoinRequest until foreman accepts requests or ROSCA has started
+   */
+  function joinRequest() onlyBeforeStart {
+    // only put the request in the pending list if they are not in the ROSCA already
+    if (members[msg.sender].alive) throw;
+    pendingJoinRequest[msg.sender] = true;
+    LogParticipantApplied(msg.sender);
+  }
 
-   /**
-    * foreman can call this method to approve a join request by a participant
-    * once a requestor had been registered as member, the address will be taken out of pendingJoinRequest
-    */
-   function acceptJoinRequest(address requestor)
-     onlyForeman
-     onlyBeforeStart {
-     if (!pendingJoinRequest[requestor]) throw;
-     addMember(requestor);
-     LogParticipantApproved(requestor);
-     delete(pendingJoinRequest[requestor]); // take out the requestor's address in the pending list
-   }
+  /**
+   * foreman can call this method to approve a join request by a participant
+   * once a requestor had been registered as member, the address will be taken out of pendingJoinRequest
+   */
+  function acceptJoinRequest(address requestor)
+    onlyForeman
+    onlyBeforeStart {
+    if (!pendingJoinRequest[requestor]) throw;
+    addMember(requestor);
+    LogParticipantApproved(requestor);
+    delete(pendingJoinRequest[requestor]); // take out the requestor's address in the pending list
+  }
 
-   /**
-    * Processes a periodic contribution from msg.sender ().
-    * Any excess funds will be withdrawable through withdraw().
-    */
-   function contribute() payable {
-     if (!members[msg.sender].alive || currentRound == 0) throw;
-     members[msg.sender].contributed += msg.value;
-     members[msg.sender].goodStanding = members[msg.sender].contributed / contributionSize;
-     if(members[msg.sender].contributed > contributionSize * membersAddresses.length) {
-       uint excessContribution = members[msg.sender].contributed - (contributionSize * membersAddresses.length);
-       members[msg.sender].contributed -= excessContribution;
-       members[msg.sender].pendingWithdrawal += excessContribution;
-     }
-     LogContributionMade(msg.sender, msg.value);
-   }
+  /**
+   * Processes a periodic contribution from msg.sender ().
+   * Any excess funds will be withdrawable through withdraw().
+   */
+  function contribute() payable {
+    if (!members[msg.sender].alive || currentRound == 0) throw;
+    members[msg.sender].contributed += msg.value;
+    members[msg.sender].goodStanding = members[msg.sender].contributed / contributionSize;
+    if(members[msg.sender].contributed > contributionSize * membersAddresses.length) {
+      uint excessContribution = members[msg.sender].contributed - (contributionSize * membersAddresses.length);
+      members[msg.sender].contributed -= excessContribution;
+      members[msg.sender].pendingWithdrawal += excessContribution;
+    }
+    LogContributionMade(msg.sender, msg.value);
+  }
 
-   /**
-    * Registers a bid from msg.sender. If msg.sender has already won a round or bid is higher than lowestBid,
-    * this method will throw.
-    */
-   function bid(uint distrubtionAmountInWei) {
-     if (distrubtionAmountInWei >= lowestBid ||
-         members[msg.sender].paid  ||
-         currentRound == 0 ||
-         members[msg.sender].goodStanding < currentRound ||
-         distrubtionAmountInWei < ((contributionSize * membersAddresses.length)/100) * MIN_DISTRIBUTION_RATIO) throw;
-     lowestBid = distrubtionAmountInWei;
-     winnerAddress = msg.sender;
-     LogNewLowestBid(lowestBid, winnerAddress);
-   }
+  /**
+   * Registers a bid from msg.sender. If msg.sender has already won a round or bid is higher than lowestBid,
+   * this method will throw.
+   */
+  function bid(uint distrubtionAmountInWei) {
+    if (distrubtionAmountInWei >= lowestBid ||
+        members[msg.sender].paid  ||
+        currentRound == 0 ||
+        members[msg.sender].goodStanding < currentRound ||
+        distrubtionAmountInWei < ((contributionSize * membersAddresses.length)/100) * MIN_DISTRIBUTION_RATIO) throw;
+    lowestBid = distrubtionAmountInWei;
+    winnerAddress = msg.sender;
+    LogNewLowestBid(lowestBid, winnerAddress);
+  }
 
-   /**
-    * Withdraws available funds for msg.sender. If opt_destination is specified,
-    * sends the fund to that address.
-    */
-   function withdraw(address opt_destination) returns(bool success) {
-     if (opt_destination == 0)
-       opt_destination = msg.sender;
-     if (!members[msg.sender].alive || members[msg.sender].pendingWithdrawal == 0 || currentRound == 0) throw;
-     return sendFund(msg.sender, opt_destination);
-   }
-
- }
+  /**
+   * Withdraws available funds for msg.sender. If opt_destination is specified,
+   * sends the fund to that address.
+   */
+  function withdraw(address opt_destination) returns(bool success) {
+    if (opt_destination == 0)
+      opt_destination = msg.sender;
+    if (!members[msg.sender].alive || members[msg.sender].pendingWithdrawal == 0 || currentRound == 0) throw;
+    return sendFund(msg.sender, opt_destination);
+  }
+}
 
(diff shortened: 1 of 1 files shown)


rmerom (approved):

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
  * The winner of the pot is decided by reverse auction (lowest Bid wins).
  *
  * Things still missing/needs fixing:
- * in withdraw() , add fee related logic
- * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+ * in withdraw() , add fee related logic (done, Sol should still be discussed)
+ *   Sol: will keep deduct a fee, and keep it in contract untill the end?
```
rmerom: sounds good to me, we might just add an entry in pendingWithdrawl for WETRUST_FEE_ADDRESS .

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ * if there are contribution than what is required, put them in pendingWithdrawal (done)
```
rmerom: s/contribution/more contributions

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
  * The winner of the pot is decided by reverse auction (lowest Bid wins).
  *
  * Things still missing/needs fixing:
- * in withdraw() , add fee related logic
- * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round
+ * in withdraw() , add fee related logic (done, Sol should still be discussed)
```
rmerom: please remove anything that is done

contracts/ROSCA.sol
rmerom: - s/untill/until
- the "Sol" words are not very clear, can we remove them?

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   Sol: will keep deduct a fee, and keep it in contract untill the end?
+ * if there are contribution than what is required, put them in pendingWithdrawal (done)
+ *   Sol: in contribute , check if contributed is enough to last the whole ROSCA, if it is put the excess in pendingWithdrawal
```
rmerom: why not put any extra in pendingWithdrawl ?

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ * when the rosca ends all the money should go out to the account user (done, still needs to work out sending fees to wetrust address)
```
rmerom: please remove anything that is done, same for next comments

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   Sol: added endOfROSCA, cleanUp and sendFUND function
+ * checkingContribution, check whether or not a member as contributed as least contributionSize at the end of Round (goodStanding var takes care of this i think)
```
rmerom: s/as/has
remove "i think" :)

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *  -startRound should check?
```
rmerom: not clear what this is?

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ * if user doesnt contribute , dont allow bidding EVER! have to be in good standing to bid/win the Pot (done)
+ *   Sol: used goodStanding variable to keep track of how many round a person can bet
+ * contribute and withdraw doesnt check if round has started (done)
+ *
+ * Product Decision to think about:
+ *   - biding when not everyone had contributed, total contributed is less than the bid amount
+ *     Possible Solution:
+ *       - owed variable? before paying the next round winner, owed is paid first?
+ *       - startRound Check ? if balance is less than lowest Bid ? if it is, what to do?
+ *       - in withdraw, check balance of contract first and if its less than pending withdrawal, subtract that amount only (I think this is the simplest option) (implement for now)
```
rmerom: working on this here: 
https://docs.google.com/document/d/1bFmwAQJNinKrrg7eaiqcBoYDVxWGm1rs2G-F4mGuAS4/edit#
Let's continue thinking about this

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *
+ * Things brought up that needs considering:
+ *   - allowing constructor to have initial list of members specified by foreman on ROSCA deployment
```
rmerom: I don't think we'll actually need that, at least at first, because usually people don't know their friends' wallet addresses

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   - using a better unit than wei to represent ether i.e finney (we shouldn't be dealing with any value less than finney anyways)
```
rmerom: we could do this, but it may be error prone because ethereum's basic units (e.g. msg.value) are in wei.
I think there is already in fact a bug in L165, where we call send() with an amount that is specified in finney.
I strongly recommend using wei as units.

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   - getters for necessary variables (making variables public would work)
```
rmerom: which vars do you think should be public? please specify.

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   - optimizing , contract deployment currently cost over 20 cents on test-net
```
rmerom: right - we discussed that back in starbucks and decided to postpone optimizing (right now we're only working on MVP)

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   - someway to kill the contract when ROSCA is ended? (currently cleanUp functions includes selfdestruct)
```
rmerom: sounds good to me

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+ *   - frontend should try a call function and if it throws, warns user of possible throw before allowing transaction.
```
rmerom: sounds good, but i'm not sure why this should be a comment in the contract

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
  */
 contract ROSCA {
-  uint32 constant MIN_CONTRIBUTION_SIZE = uint32(1000000000000);  // 1e12
-  uint32 constant MAX_CONTRIBUTION_SIZE = uint32(10000000000000000000); // 10 ether in Wei
-  uint32 constant MAX_FEE_IN_THOUSANDTHS = 200;
-  address constant WETRUST_FEE_ADDRESS = 0x0;           // TODO: needs to be updated
+  uint64 constant MIN_CONTRIBUTION_SIZE = 1 finney;  // 1e12
```
rmerom: nice!
s/1e12/1e12 wei

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+  uint64 constant MAX_CONTRIBUTION_SIZE = 10 ether;//uint64(10000000000000000000); // 10 ether in Wei
```
rmerom: no need for the comment (10 ether is pretty clear)

rmerom: uint64 can hold about 10^20. 
10^20 wei == 100 ether, which future ROSCAs may cross later on.

Please change to uint128.

contracts/ROSCA.sol
```diff
@@ -6,19 +6,43 @@ pragma solidity ^0.4.4;
...
+  uint16 constant MAX_FEE_IN_THOUSANDTHS = 200;
   uint32 constant MINIMUM_TIME_BEFORE_ROSCA_START = 1 days;   // startTime of the ROSCA must be at least 1 day away from when the ROSCA is created
-  uint32 constant MINIMUM_PARTICIPANTS = 2;           // minimum participants for the ROSCA to start
-  uint32 constant MIN_ROUND_PERIOD_IN_DAYS = 1;
-  uint32 constant MAX_ROUND_PERIOD_IN_DAYS = 30;
-  uint32 constant MIN_DISTRIBUTION_RATIO = 65;  // the winning bid must be at least 65% of the Pot value
+  uint8 constant MINIMUM_PARTICIPANTS = 2;           // minimum participants for the ROSCA to start
+  uint8 constant MIN_ROUND_PERIOD_IN_DAYS = 1;
+  uint8 constant MAX_ROUND_PERIOD_IN_DAYS = 30;
+  uint8 constant MIN_DISTRIBUTION_RATIO = 65;  // the winning bid must be at least 65% of the Pot value
```
rmerom: this is not really RATIO, it's actually PERCENT. Please update var name.

contracts/ROSCA.sol
```diff
@@ -33,27 +57,27 @@ contract ROSCA {
   uint16 serviceFeeInThousandths;
   uint16 currentRound;  // currentRound will be set to 0 when ROSCA is created and will turn to one when the ROSCA actually starts
   uint16 minParticipants;
+  bool endOfROSCA = false;
   address foreman;
-  uint contributionSize;
+  uint32 contributionSize;
```
rmerom: I guess you changed this because of switching to finney. I strongly suggest we stay with wei because of error-proneness mentioned above, in which case please leave as uint or uint128.

contracts/ROSCA.sol
```diff
@@ -69,32 +93,28 @@ contract ROSCA {
     */
   function ROSCA(
     uint16 roundPeriodInDays_,
-    uint contributionSize_,
+    uint32 contributionSize_,
```
rmerom: same as above, in case we switch to wei, let's change this back to uint or uint128.

contracts/ROSCA.sol
```diff
@@ -69,32 +93,28 @@ contract ROSCA {
...
     uint16 minParticipants_,
     uint startTime_,
     uint16 serviceFeeInThousandths_) {
-    if (roundPeriodInDays < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodInDays > MAX_ROUND_PERIOD_IN_DAYS) throw;
+    if (roundPeriodInDays_ < MIN_ROUND_PERIOD_IN_DAYS || roundPeriodInDays_ > MAX_ROUND_PERIOD_IN_DAYS) throw;
```
rmerom, line 72: good catch!

contracts/ROSCA.sol
```diff
@@ -69,32 +93,28 @@ contract ROSCA {
...
     roundPeriodInDays = roundPeriodInDays_;
     if (contributionSize_ < MIN_CONTRIBUTION_SIZE || contributionSize_ > MAX_CONTRIBUTION_SIZE) throw;
     contributionSize = contributionSize_;
 
-
     if (minParticipants_ < MINIMUM_PARTICIPANTS) throw; // there should be at least 2 people to make a group
     minParticipants = minParticipants_;
-
-    if (startTime_ < (now + MINIMUM_TIME_BEFORE_ROSCA_START)) throw;
+     if (startTime_ < (now + MINIMUM_TIME_BEFORE_ROSCA_START)) throw;
```
rmerom: wrong indentation

contracts/ROSCA.sol
```diff
@@ -69,32 +93,28 @@ contract ROSCA {
...
     startTime = startTime_;
-
-    if (serviceFeeInThousandths_ > MAX_FEE_IN_THOUSANDTHS) throw;
+     if (serviceFeeInThousandths_ > MAX_FEE_IN_THOUSANDTHS) throw;
```
rmerom: wrong indentation

contracts/ROSCA.sol
```diff
@@ -69,32 +93,28 @@ contract ROSCA {
...
     serviceFeeInThousandths = serviceFeeInThousandths_;
 
     foreman = msg.sender;
-
     addMember(msg.sender);
   }
 
   function addMember(address newMember) internal {
-    members[newMember] = User({paid: false , contributed: 0, alive: true, pendingWithdrawl: 0});
+    members[newMember] = User({paid: false , contributed: 0, alive: true, pendingWithdrawal: 0,goodStanding: 0});
```
rmerom: space after comma

contracts/ROSCA.sol
```diff
@@ -120,7 +140,7 @@ contract ROSCA {
           break;
         }
       }
-      members[winnerAddress].pendingWithdrawl += lowestBid;
+      members[winnerAddress].pendingWithdrawal += (lowestBid / 100000) * MAX_FEE_IN_THOUSANDTHS;
```
rmerom: if changing back to wei, please also revert this

contracts/ROSCA.sol
```diff
@@ -120,7 +140,7 @@ contract ROSCA {
           break;
```
rmerom, line 112: in L134 (I can't comment there for some reason) the comment "only true when there is no bidder in this round" is redundant because the next line says the same thing

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
 
       currentRound++;
       LogStartOfRound(currentRound);
+    } else if (currentRound == membersAddresses.length) {
```
rmerom: do we need this "if" or can we only leave the "else" in place?

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
+        endOfROSCA = true;
     }
   }
 
   /**
+   * try to send the fund to the opt_destination from source account
+   */
+  function sendFund(address source,address opt_destination) internal returns(bool success){
+    uint amountToWithdraw = members[source].pendingWithdrawal  ;  // use a temporary variable to avoid the receiver calling the withdraw function again before
+    if(this.balance < amountToWithdraw) amountToWithdraw = this.balance;
```
rmerom: space before (

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
+    members[source].pendingWithdrawal = members[source].pendingWithdrawal - amountToWithdraw;                      // send() complete its task
```
rmerom: I don't understand the comment?

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
+    if (!opt_destination.send(amountToWithdraw)) {   // if the send() fails, put the allowance back to its original place
```
rmerom: i think there is a bug here: send() expects amount in wei. 
amountToWithdraw is in finney. Let's switch to using wei everywhere to avoid such mistakes.

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
+      // No need to call throw here, just reset the amount owing
+      members[source].pendingWithdrawal = amountToWithdraw;
```
rmerom: possibly a bug: I think this should be
members[source].pendingWithdrawal += amountToWithdraw;

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
+      return true;
+    }
+  }
+
+  /**
+   * cleanUp is to make sure that fees are send to where they are owed and
+   * make sure contract doesn't hold any ether after it ends.
+   *
+   */
+  function cleanUp() {
```
rmerom, line 151: hmmm what is the incentive for anyone to call this? I'm afraid no one will want to waste gas on that.
Moreover, since participants' addresses may be contracts themselves, this call could take a LOT of gas.
I suggest removing this function, letting participants draw their own money, and letting WeTrust draw its money (we'll have the frontend do that). This way no one needs to worry about gas of other people's contracts.

contracts/ROSCA.sol
```diff
@@ -162,8 +219,14 @@ contract ROSCA {
    * Any excess funds will be withdrawable through withdraw().
    */
   function contribute() payable {
-    if (!members[msg.sender].alive) throw;
+    if (!members[msg.sender].alive || currentRound == 0) throw;
```
rmerom: you might want to also add a throw if endOfROSCA

contracts/ROSCA.sol
```diff
@@ -162,8 +219,14 @@ contract ROSCA {
...
     members[msg.sender].contributed += msg.value;
+    members[msg.sender].goodStanding = members[msg.sender].contributed / contributionSize;
```
rmerom: I think the name goodStanding is a bit misleading - this is the number of rounds a participant has fully contributed to.
However, this is easily computed from members[msg.sender].contributed, so why do we need to store it in another var?

contracts/ROSCA.sol
```diff
@@ -162,8 +219,14 @@ contract ROSCA {
...
+    if(members[msg.sender].contributed > contributionSize * membersAddresses.length) {
```
rmerom: space after if

contracts/ROSCA.sol
```diff
@@ -162,8 +219,14 @@ contract ROSCA {
...
+      uint excessContribution = members[msg.sender].contributed - (contributionSize * membersAddresses.length);
+      members[msg.sender].contributed -= excessContribution;
```
rmerom, line 189: This is correct logic, but I think our code is getting too complex with "contributed", "goodStanding" and "pendingWithdrawl".

What if we do the following: we rename "pendingWithdrawl" to "credit" and make it a signed int.
* When someone sends a contribution, we add it to "credit"
* At end of round, we deduct contributionSize from the "credit" as many participants as possible, and add lowestBid to the winner's "credit"
* When someone wants to bid, we first validate that they have nonnegative "credit"
* When someone wants to withdraw, they can withdraw up to "credit" (or zero if negative)

What do you think?

contracts/ROSCA.sol
```diff
@@ -175,6 +238,7 @@ contract ROSCA {
     if (distrubtionAmountInWei >= lowestBid ||
```
rmerom: can we rename "distributionAmountInWei" to "bidInWei" for clarity (as we're comparing to lowestBid later)?

contracts/ROSCA.sol
```diff
@@ -130,10 +150,47 @@ contract ROSCA {
...
       LogStartOfRound(currentRound);
+    } else if (currentRound == membersAddresses.length) {
+        endOfROSCA = true;
     }
   }
 
   /**
+   * try to send the fund to the opt_destination from source account
+   */
+  function sendFund(address source,address opt_destination) internal returns(bool success){
```
rmerom: 1. I've used the prefix "opt_" to specify that an argument is optional. Here I don't think it's optional, so can we rename it to "destination".

2. I think the name "source" is misleading because it implies that funds come from it, whereas the funds come from the contract itself. Consider renaming to "participant" and updating the comment accordingly.

contracts/ROSCA.sol
shine2lay: There is a reason i brought this up, my tests shows that we are nearing the gas limit with the contract as it stands, if we dont want to optimize, we'll have to break up contract into multiple contracts

contracts/ROSCA.sol
shine2lay: i like to write down what i m thinking, it was just convenient at the time,

contracts/ROSCA.sol
shine2lay: because its a lot less efficient if the person wants to contribute enough for the whole ROSCA up front or even just a couple Rounds worth of contributions

contracts/ROSCA.sol
shine2lay: just my thought process, removed

contracts/ROSCA.sol
shine2lay: Hoang and Tom brought it up, we can discuss it more tonight

contracts/ROSCA.sol
shine2lay: well, it is in wei, 1 finney = 1e14 wei i.e MIN_CONTRIBUITION_SIZE = 1 finney will be stored as 1e14 wei. It should provide better readability overall.

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: changed to uint128

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: removed

contracts/ROSCA.sol
shine2lay: don't want to take chances in case of attacker managed to make currentRound > membersAddresses.length

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: my thought process , removed

contracts/ROSCA.sol
shine2lay: everything is in wei

contracts/ROSCA.sol
shine2lay: yes , done

contracts/ROSCA.sol
shine2lay: destroying storage and contracts refunds gas and it makes sure there is no fund stuck in contract at the end

contracts/ROSCA.sol
shine2lay: point taken, removed

contracts/ROSCA.sol
shine2lay: i think its a lot more work and a ton of wasted gas because you'll have to go through all the participants in a loop and it will be very expensive. I mean very, you need to STORAGEMOD for every participants and they are not cheap.

I took out goodStanding, added one more check to bidding, which essentially replaces goodStanding

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: something George and Patrick brought up, something to make sure there is no ether stuck in contract

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: those are mainly to jot my memory of how i went about trying/planning to solve the issue

contracts/ROSCA.sol
shine2lay: havent given much thought, on the top of my head, currentRound, lowestBid, members

contracts/ROSCA.sol
shine2lay: ok

contracts/ROSCA.sol
shine2lay: thanks?

contracts/ROSCA.sol
shine2lay: ok

shine2lay:
Review status: 0 of 1 files reviewed at latest revision, 25 unresolved discussions.

---

*[contracts/ROSCA.sol, line 11 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-11:-KXJtqdfg0I7Hkbk9n2t:brpmqb3) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L11)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> s/contribution/more contributions
</details>
removed

---

*[contracts/ROSCA.sol, line 143 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-143:-KXJu7QwHKnZdAgh8qqe:bec7v69) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L143)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> if changing back to wei, please also revert this
</details>
don't have to , its always been in Wei

---

*[contracts/ROSCA.sol, line 225 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-225:-KXJuFqkMoQBn1nI7G1i:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L225)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> space after if
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

shine2lay:
Review status: 0 of 1 files reviewed at latest revision, 23 unresolved discussions.

---

*[contracts/ROSCA.sol, line 13 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-13:-KXJuU0BIatGKqmP5YAf:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L13)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> please remove anything that is done, same for next comments
</details>
Done.

---

*[contracts/ROSCA.sol, line 26 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-26:-KXJuVJZgN94Qkauc8rf:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L26)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> working on this here: 
> 
> https://docs.google.com/document/d/1bFmwAQJNinKrrg7eaiqcBoYDVxWGm1rs2G-F4mGuAS4/edit#
> 
> Let's continue thinking about this
</details>
Done.

---

*[contracts/ROSCA.sol, line 107 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-107:-KXJuWYBEXhUk8kZMGXR:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L107)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> wrong indentation
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

rmerom:
Review status: 0 of 1 files reviewed at latest revision, 17 unresolved discussions.

---

*[contracts/ROSCA.sol, line 9 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-9:-KXJqJshpezRrK1vAR7r:b1kkmco) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L9)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> done
</details>
i'm still seeing "(done, Sol should still be discussed)". Is there a newer version?

---

*[contracts/ROSCA.sol, line 10 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-10:-KXJqTA8Y4tqAjGS7ZUi:b31q0lz) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L10)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> those are mainly to jot my memory of how i went about trying/planning to solve the issue 
</details>
I see, do you think we can remove that before pushing, and keep tasks elsewhere, e.g. in trello?
also, please s/untill/until

---

*[contracts/ROSCA.sol, line 12 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-12:-KXJiA4iDdtPcQfI4muv:bkvjqz6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L12)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> because its a lot less efficient if the person wants to contribute enough for the whole ROSCA up front or even just a couple Rounds worth of contributions 
</details>
I mean using pendingWithdrawl as a "credit" var, as described in https://docs.google.com/document/d/1bFmwAQJNinKrrg7eaiqcBoYDVxWGm1rs2G-F4mGuAS4/edit

---

*[contracts/ROSCA.sol, line 30 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-30:-KXJsFQ4DxoXHgM5Mws-:b-hfe2k7) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L30)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> well, it is in wei, 1 finney = 1e14 wei i.e MIN_CONTRIBUITION_SIZE = 1 finney will be stored as 1e14 wei. It should provide better readability overall. 
</details>
I see. I misunderstood (I thought you meant holding number of finney in each var). SGTM.

---

*[contracts/ROSCA.sol, line 31 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-31:-KXJsPUGsWxPsOa5JRKN:bmmikuw) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L31)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> havent given much thought, on the top of my head, currentRound, lowestBid, members
</details>
I find that keeping the code clean and putting comments and thoughts elsewhere (e.g. in Google Docs, Trello etc.)  really helps the readability and auditability of the code. These preliminary thoughts are really important to write down so we don't forget them, just maybe we find a better place for them?

---

*[contracts/ROSCA.sol, line 32 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-32:-KXJsxCmaZ6BqavAH9Mc:b3gq65m) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L32)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> There is a reason i brought this up, my tests shows that we are nearing the gas limit with the contract as it stands, if we dont want to optimize, we'll have to break up contract into multiple contracts
</details>
i see, that's important. could you share (not necessarily here, email is also good) the calls and the gas involved?

---

*[contracts/ROSCA.sol, line 34 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-34:-KXJt7-nbLW-tTf_qLqN:bukmx34) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L34)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> i like to write down what i m thinking, it was just convenient at the time, 
</details>
i get that. do you think we can move these thoughts to other places before pushing the code?

---

*[contracts/ROSCA.sol, line 153 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-153:-KXJu0QKnMTJ24S8P1m1:b-frzvqb) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L153)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> don't want to take chances in case of attacker managed to make currentRound > membersAddresses.length
</details>
I think in this case, even if an attacker manages to somehow get currentRound to be greater than membersAddresses.length, the only reprecussion is that endOfRosca will be assigned true (not that bad considering we have a major bug).
As extra instructions cost gas (and code complexity), maybe we should keep extra security to where the implications are major, e.g. sending money?

---

*[contracts/ROSCA.sol, line 163 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-163:-KXKBxQuWiONuUhRc6oa:b-oqm5ib) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L163)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> done
</details>
i can't see that; can you double check?

---

*[contracts/ROSCA.sol, line 167 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-167:-KXKC7qhzWI1WTnIrqGd:bs4jvz9) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L167)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> yes , done 
</details>
my mistake; this should be -= , not +=

---

*[contracts/ROSCA.sol, line 180 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-180:-KXKCDsz9Czi2y3PsSYb:b-v1hnun) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L180)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> something George and Patrick brought up, something to make sure there is no ether stuck in contract
</details>
I'm not sure what you and George/Patick mean by "money stuck in contract". If it's money remaining due to a bug, then this mechanism would channel unused-by-bug funds to WeTrust, which seems unfair.

If it's people not claiming their credits, we should let them do it. The receiving address may belong to a contract, meaning the send() call can be costly. Having one function call send()s for all participants is even more costly and may either surpass the gas limit, or at least have the caller pay a big tx fee which is unfair. 
It's valid concern not to let people leave money in the contract, but we can have the frontend them to withdraw these funds, which would make more sense.

Re refunds, they are deducted from the calling transaction gas fees, so that in itself doesn't seem to be  an incentive for anyone to call cleanUp().

Do you feel we should have cleanUp()? if so, can you explain why?

---

*[contracts/ROSCA.sol, line 222 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-222:-KXKDVxFe3pyedNRNNO4:b941x6h) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L222)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> ok
</details>
I can't see this change?

---

*[contracts/ROSCA.sol, line 227 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-227:-KXKDtAccRwfqnVaHVwY:b-hhsbhz) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L227)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> i think its a lot more work and a ton of wasted gas because you'll have to go through all the participants in a loop and it will be very expensive. I mean very, you need to STORAGEMOD for every participants and they are not cheap.
> 
> 
> 
> I took out goodStanding, added one more check to bidding, which essentially replaces goodStanding
</details>
You're right, updating all participants whenever there is a winner is not a good idea due to gas wastage. However, keeping both contributed and pendingWithdrawl is tricky, as they are not normalized (they depend on each other). 
There are more available approaches - let's continue this discussion on https://docs.google.com/document/d/1bFmwAQJNinKrrg7eaiqcBoYDVxWGm1rs2G-F4mGuAS4/edit# 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

shine2lay:
Review status: 0 of 1 files reviewed at latest revision, 17 unresolved discussions.

---

*[contracts/ROSCA.sol, line 9 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-9:-KXNpusdi0w-3BFjg5vd:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L9)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> i'm still seeing "(done, Sol should still be discussed)". Is there a newer version?
</details>
Done.

---

*[contracts/ROSCA.sol, line 10 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-10:-KXNpycioefG1wZiy1Js:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L10)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I see, do you think we can remove that before pushing, and keep tasks elsewhere, e.g. in trello?
> also, please s/untill/until
</details>
Done.

---

*[contracts/ROSCA.sol, line 30 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-30:-KXNl9CG_ikyRdPcgouT:bb73djm) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L30)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I see. I misunderstood (I thought you meant holding number of finney in each var). SGTM.
</details>
SGTM

---

*[contracts/ROSCA.sol, line 31 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-31:-KXNlEdYY0Mo53bJ9oh_:b92v2wh) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L31)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I find that keeping the code clean and putting comments and thoughts elsewhere (e.g. in Google Docs, Trello etc.)  really helps the readability and auditability of the code. These preliminary thoughts are really important to write down so we don't forget them, just maybe we find a better place for them?
</details>
personally i like having them especially in through reviewing process, it not only helps me keep track of what i did in this particular version. It also would provide how this version is changed compare to the previous one. I agree that we need to take it out before merging, i'll take it out in the version that is ready to be merge. 

---

*[contracts/ROSCA.sol, line 32 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-32:-KXNm0LwwL2LFSQvSSEn:b-qivnso) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L32)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> i see, that's important. could you share (not necessarily here, email is also good) the calls and the gas involved?
</details>
It is no longer necessary , if we take out request Join system, we should have plenty of gas limit left to add

---

*[contracts/ROSCA.sol, line 34 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-34:-KXNly5qsQNg5HGHhWvU:bihedew) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L34)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> i get that. do you think we can move these thoughts to other places before pushing the code?
</details>
read the comment above

---

*[contracts/ROSCA.sol, line 153 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-153:-KXNq0ga98N2HAxC9_T4:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L153)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I think in this case, even if an attacker manages to somehow get currentRound to be greater than membersAddresses.length, the only reprecussion is that endOfRosca will be assigned true (not that bad considering we have a major bug).
> As extra instructions cost gas (and code complexity), maybe we should keep extra security to where the implications are major, e.g. sending money?
> 
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

shine2lay:
Review status: 0 of 1 files reviewed at latest revision, 17 unresolved discussions.

---

*[contracts/ROSCA.sol, line 163 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-163:-KXRWlfGJcLSqFppzQl1:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L163)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> i can't see that; can you double check?
</details>
Done.

---

*[contracts/ROSCA.sol, line 167 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-167:-KXRWmtwp2bSDOKd9Tzv:bk9qava) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L167)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> my mistake; this should be -= , not +=
</details>
it should be += , its added the amountToWithdraw back into pendingWithdrawal in case when .send() fails
 

---

*[contracts/ROSCA.sol, line 180 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-180:-KXRYZ3VrtXNAnhItPGd:b2um5co) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L180)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I'm not sure what you and George/Patick mean by "money stuck in contract". If it's money remaining due to a bug, then this mechanism would channel unused-by-bug funds to WeTrust, which seems unfair.
> 
> If it's people not claiming their credits, we should let them do it. The receiving address may belong to a contract, meaning the send() call can be costly. Having one function call send()s for all participants is even more costly and may either surpass the gas limit, or at least have the caller pay a big tx fee which is unfair. 
> It's valid concern not to let people leave money in the contract, but we can have the frontend them to withdraw these funds, which would make more sense.
> 
> Re refunds, they are deducted from the calling transaction gas fees, so that in itself doesn't seem to be  an incentive for anyone to call cleanUp().
> 
> Do you feel we should have cleanUp()? if so, can you explain why?
> 
> 
> 
> 
</details>
Even if we take out the sending ether part, we should still keep a way to kill the contract after ROSCA, which is always good for the blockchain.

---

*[contracts/ROSCA.sol, line 222 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-222:-KXRWcLGj33KwT7rs5kC:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L222)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I can't see this change?
</details>
Done.

---

*[contracts/ROSCA.sol, line 227 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-227:-KXRWjw_AqV3GJs-eLEB:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L227)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> You're right, updating all participants whenever there is a winner is not a good idea due to gas wastage. However, keeping both contributed and pendingWithdrawl is tricky, as they are not normalized (they depend on each other). 
> There are more available approaches - let's continue this discussion on https://docs.google.com/document/d/1bFmwAQJNinKrrg7eaiqcBoYDVxWGm1rs2G-F4mGuAS4/edit# 
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

rmerom:
Review status: 0 of 1 files reviewed at latest revision, 6 unresolved discussions.

---

*[contracts/ROSCA.sol, line 29 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-29:-KXUoJJpjz1tH4k3LW_1:bmeyrwp) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L29)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> Hoang and Tom brought it up, we can discuss it more tonight
</details>
:+1: 

---

*[contracts/ROSCA.sol, line 31 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-31:-KXUoUI1bTnzNM0oRZzA:b-eucyc1) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L31)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> personally i like having them especially in through reviewing process, it not only helps me keep track of what i did in this particular version. It also would provide how this version is changed compare to the previous one. I agree that we need to take it out before merging, i'll take it out in the version that is ready to be merge. 
</details>
Yes, that makes a lot of sense. In future, maybe a solution that can both help us keep track of the changes, and keep the code clean, is putting the changes in the PR description. 
It'll also help us in future find easily where we introduced a certain change to the code. Would that work for you?

---

*[contracts/ROSCA.sol, line 167 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-167:-KXUpgfEIzQ7VYzcQ1MF:b-29kifc) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L167)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> it should be += , its added the amountToWithdraw back into pendingWithdrawal in case when .send() fails
>  
</details>
right.. i need more sleep :face_with_head_bandage: 

---

*[contracts/ROSCA.sol, line 180 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-180:-KXUq-SNdqeYYasWIIhe:b8n5uam) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L180)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> Even if we take out the sending ether part, we should still keep a way to kill the contract after ROSCA, which is always good for the blockchain.
</details>
I like this direction, but I think the current solution still has issues: as it is now, right after the last ROSCA round ends, the foreman can collect all funds, even before participants get a chance to do it. We can allow the foreman to do that maybe some FOREMAN_RETRIEVE_LEFTOVER_FUNDS_DELAY time after the ROSCA ends. WDYT?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

shine2lay:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/ROSCA.sol, line 29 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-29:-KXWcs2f0_db7rS_QGfj:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L29)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> :+1: 
</details>
Done.

---

*[contracts/ROSCA.sol, line 31 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-31:-KXWcwdkvyDSBTYp_MQ7:bumclj6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L31)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> Yes, that makes a lot of sense. In future, maybe a solution that can both help us keep track of the changes, and keep the code clean, is putting the changes in the PR description. 
> It'll also help us in future find easily where we introduced a certain change to the code. Would that work for you?
</details>
it'll be hard to put all the changes in the PR since each version deals with more than one thing, here is what i propose, first iteration of the review , i want to leave the comments in as it would also help with tracking which version would of the files contain the changes and easy to find the revision that corresponds to it in case if we needs to go back to a certain change, after the first iteration, i'll take it out 

---

*[contracts/ROSCA.sol, line 167 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-167:-KXWde0FMyv1-9gUNUlO:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L167)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> right.. i need more sleep :face_with_head_bandage: 
</details>
Done.

---

*[contracts/ROSCA.sol, line 180 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-180:-KXWdlPlTIo1x-Mf7-PU:beyvpto) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L180)):*
<details><summary><i>Previously, rmerom (Ron Merom) wrote…</i></summary>
> I like this direction, but I think the current solution still has issues: as it is now, right after the last ROSCA round ends, the foreman can collect all funds, even before participants get a chance to do it. We can allow the foreman to do that maybe some FOREMAN_RETRIEVE_LEFTOVER_FUNDS_DELAY time after the ROSCA ends. WDYT?
</details>
thats my original thought, we should set aside maybe 2 Rounds worth of time after ROSCA ends untill Foreman can call cleanUp WDYT?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

rmerom:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[contracts/ROSCA.sol, line 180 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7#-KXJfto26PkaDhRvzpmF-r1-180:-KXWhonOmuFQE2dXUVwY:bug9rn2) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/4c8cfd1d8de2932ae0fca6d5489297776dbbfd48/contracts/ROSCA.sol#L180)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary>
> thats my original thought, we should set aside maybe 2 Rounds worth of time after ROSCA ends untill Foreman can call cleanUp WDYT?
</details>
ok i see that we only allow cleanUp if there is no more money in the contract, then it's not really necessary to wait.
i thought this was to allow the foreman to collect any leftover funds (and give them to participants offline)? if not, no problem, we can leave it as is. I'm good with whatever you decide.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

rmerom:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/7)*

PR #11, opened 2016-12-02
publicizer - a tool to replace internal and private methods and vars with public for testing purposes

rmerom:
Related to our previous discussion, the new tool "publicizer" replaces
```  uint256 internal /* publicForTesting */ someVar;```
with
```  uint256 public /* modifiedForTest */ someVar;```

This is useful to let tests gain access to internal state variables and methods of the production contract.

We could have replaced every single instance of "internal" or "private" with "public" but this takes a more conservative approach where we specify explicitly the vars and functions we want to change, and we mark them also in the test contract.

**Usage**:
```publicizer.py contract.sol > contract_test.sol```

It would be nice later on to integrate it directly into ```truffle test```.

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/11)

### DIFF
--- a/tools/build/publicizer.py
+++ b/tools/build/publicizer.py
@@ -1,24 +1,30 @@
 #!/usr/bin/env python3
-import fileinput
+
+"""
+Usage: publicizer.py someFile.sol
+
+Replaces
+  uint256 internal /* publicForTesting */ someVar;
+with
+  uint256 public /* modifiedForTest */ someVar;
+
+Also replaces "private" instead of "internal", as long as they're followed by
+/* publicForTesting */
+
+The modified file is printed to stdout.
+
+"""
+
 import re
 import sys
 
-# Usage: publicizer.py someFile.sol
-#
-# Replaces
-#   uint256 internal /* publicForTesting */ someVar;
-# with
-#   uint256 public someVar;
-#
-# Also replaces "private" instead of "internal", as long as they're followed by
-# /* publicForTesting */
-#
-# The modified file is printed to stdout.
-
-p = re.compile('\s+(internal|public)\s*/\*\s*publicForTesting\s*\*/\s*')
-replaceWith = ' public /* modifiedForTest */ '
 
 def substitute(content):
+  """Does the actual substitution"""
+
+  p = re.compile(r'\s+(internal|public)\s*/\*\s*publicForTesting\s*\*/\s*')
+  replaceWith = ' public /* modifiedForTest */ '
+
   return p.sub(replaceWith, content)
 
 if __name__ == "__main__":
--- a/tools/build/publicizer.py
+++ b/tools/build/publicizer.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+import fileinput
+import re
+import sys
+
+# Usage: publicizer.py someFile.sol
+#
+# Replaces
+#   uint256 internal /* publicForTesting */ someVar;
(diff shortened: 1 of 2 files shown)


shine2lay:
how are we going to take care of days to seconds issue?  Are we going to let front end deal with it? and also, if the contract file name is different, we'll need to change contract + constructor name into "new file name" otherwise truffle wont compile them

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KY1w7JurpG40QjuHUXp:bihsc0y)*

rmerom:
good points.
we can either leave the contract API as is, and control time explicitly (e.g. using ```eth_increaseTime``` - see slack for a link). Or we can change the contract API to work with seconds. Whatever you think is more appropriate.

Re the contract name, I'm hoping we can minimize the difference between the production contract and the contract that we're testing, as each difference slightly increases the chance we're testing something other than what's running out there. 
I wonder if we can have a script that runs the publicizer and puts the output contract, with the new name, in a different directoy, where the test picks it up. We should research more whether truffle can do something like this.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KY318lRy2p3spTygm6X:bp6tyof)*

shine2lay:
After researching evm_increaseTime, i am fairly certain, it is ok to leave the days as is. 

re contract name, we only have to change to contract name and constructor name, i think we can add that in the script you wrote. I can write a shell script that run publicizer before truffle test but that wouldn't be ideal for different OS. 

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KY4cSMxz6Wdb-xPBUW0:b-s1nmwc)*

rmerom:
PTAL. 
If you find this script may not to be very convenient, as it overwrites the original file, we can always switch to the inheritance model.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KYCsJRYBwOCHpHJvV5i:b8ah3qs)*

shine2lay:
Ben told me that it's a good idea to test the internal functions individually as well, so i ended up writing test cases for them as well, will this script take "internal" function and replace them with empty strings?

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KYplku9vNMvrsP9KZhV:bpa35ew)*

rmerom:
yes, whenever it sees the pattern *internal /* publicForTesting */*, it
will replace it

shine2lay:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/11#-:-KYud1aH_vH0skpEZE3f:bnfp4nl)*

shine2lay (approved):

PR #28, opened 2016-12-20
Added contributeUnitTests.js and withdrawUnitTest.js

shine2lay:
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/28)


rmerom:
Really great tests!

---

Reviewed 2 of 4 files at r1, 2 of 4 files at r2.
Review status: all files reviewed at latest revision, 35 unresolved discussions.

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGWyI_PYjlDy0K369:-KZPGWyI_PYjlDy0K36A:b-85823f) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*
> ```JavaScript
>         var rosca = yield ROSCATest.new(ROUND_PERIOD_IN_DAYS, CONTRIBUTION_SIZE, DayFromNow, MEMBER_LIST, SERVICE_FEE);
> 
>         return rosca.contribute({from: accounts[4], value: CONTRIBUTION_SIZE}).then(function() {
> ```

do we really need the "return" here?

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGpV_GwI3kRNKZBd5:-KZPGpV_GwI3kRNKZBd6:bhqyvij) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*
> ```JavaScript
>         var rosca = yield ROSCATest.new(ROUND_PERIOD_IN_DAYS, CONTRIBUTION_SIZE, DayFromNow, MEMBER_LIST, SERVICE_FEE);
> 
>         return rosca.contribute({from: accounts[4], value: CONTRIBUTION_SIZE}).then(function() {
> ```

I suggest that we also run a valid contribute() in the same test (e.g. from accounts[1]) to make sure we're catching the case that a non-member is contributing, and we don't get a throw because of another reason (e.g. if the production code later changes not to allow contribute() before ROSCA starts)

---

*[test/contributeUnitTest.js, line 26 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPH4EvqZu6lgE8ogNI:-KZPH4EvqZu6lgE8ogNJ:b-dbz4j7) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L26)):*
> ```JavaScript
>     }));
> 
>     it("Testing for event LogContributionMade()", co(function *() {
> ```

consider
it("generates a LogContributionMade event after a successful event")

---

*[test/contributeUnitTest.js, line 37 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPHLd4pWj2PcJ6hsOv:-KZPHLd4pWj2PcJ6hsOw:b-wp0q31) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L37)):*
> ```JavaScript
>             contributionMadeEvent.stopWatching();
>             assert.equal(log.args.user, accounts[1], "LogContributionMade doesn't display proper user value");
>             assert.equal(log.args.amount, CONTRIBUTION_SIZE * 0.1, "LogContributionMade doesn't display proper amount value");
> ```

nit: plz put the actual contribution (CONTRIBUTION_SIZE * 0.1) in a constant, so that it's easy to understand why we're expecting a certain amount.

---

*[test/contributeUnitTest.js, line 54 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPHqi_F65ksVEZQYVO:-KZPHqiajnnMEzcKsc_M:bfk2qub) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L54)):*
> ```JavaScript
>         yield rosca.contribute({from: accounts[2], value: CONTRIBUTION_SIZE});
>         var credit_after = yield rosca.members.call(accounts[2]);
>         assert.equal(credit_after[0], CONTRIBUTION_SIZE, "contribution's credit value didn't get registered properly");
> ```

I suggest contributing twice, and seeing that the credit is the sum of both contributions.
Otherwise, the following will also pass this test:
```members[msg.sender].credit = msg.value;```
( = instead of += )
```


---

*[test/withdrawUnitTest.js, line 3 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPI9nyuiDOeTZxTxcw:-KZPI9nyuiDOeTZxTxcx:b-i2tlkj) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L3)):*
> ```JavaScript
> var Promise = require("bluebird");
> var co = require("co").wrap;
> contract('ROSCA withdraw Unit Test', function(accounts) {
> ```

nit: would be nice to consistenly add a blank line above contract()

---

*[test/withdrawUnitTest.js, line 6 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIFgCnxHgkTfscTAU:-KZPIFgCnxHgkTfscTAV:bv6hof2) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L6)):*
> ```JavaScript
>     const MIN_START_DELAY = 86400 + 60;
>     const ROUND_PERIOD_DELAY = 86400 * 3;
>     const MEMBER_COUNT = 4;
> ```

nit: in order to normalize the code, consider 
const MEMBER_COUNT = MEMBER_LIST.length + 1;

(make it less likely for both these vars to fall out of sync in future)

---

*[test/withdrawUnitTest.js, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIX8klzwNnRIIB0-H:-KZPIX8klzwNnRIIB0-I:b12ln7f) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L21)):*
> ```JavaScript
>         var rosca = yield ROSCATest.new(ROUND_PERIOD_IN_DAYS, CONTRIBUTION_SIZE, DayFromNow, MEMBER_LIST, SERVICE_FEE);
> 
>         return rosca.withdraw({from: accounts[4]}).then(function() {
> ```

I suggest to first contribute from all members, so it's clear that even in that case withdraw() still throws

---

*[test/withdrawUnitTest.js, line 22 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPEXZuW2qzOyUe3-X2-r2-22:-KZPIRSV1J5iso4Xuhs0:b-qyahzj) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L22)):*
<details><summary><i>Previously, houndci-bot (Hound) wrote…</i></summary><blockquote>

'assert' is not defined.
</blockquote></details>

Please add the lines to remove this warning

---

*[test/withdrawUnitTest.js, line 36 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIentQPV4yhzzGm5x:-KZPIentQPV4yhzzGm5y:b2psdzh) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L36)):*
> ```JavaScript
>         const CONTRIBUTE_PERCENT = 0.8;
> 
>         yield rosca.contribute({from: accounts[0], value: CONTRIBUTION_SIZE * CONTRIBUTE_PERCENT});
> ```

same as above, plz put the exact contribution size in a constant, so the test is easier to read

---

*[test/withdrawUnitTest.js, line 80 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKWxqZmDkqeHucDvy:-KZPKWxqZmDkqeHucDvz:b-4g2e1i) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L80)):*
> ```JavaScript
>     }));
> 
>     it("Watch for LogCannotWithdrawFully when calling withdraw with amountToWithdraw > contract's balance", co(function *() {
> ```

amountToWithdraw is an implementation detail. Perhaps we can have the test name reflect the behavior, e.g.
it("generates a LogCannotWithdrawFully when the contract balance is less than what the user is entitled to")

---

*[test/withdrawUnitTest.js, line 111 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPJqrOZYblmfbGDr1o:-KZPJqrOZYblmfbGDr1p:bxc2ly5) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L111)):*
> ```JavaScript
>         var credit_before = yield rosca.members.call(accounts[2]);
>         var withdrawalEvent = rosca.LogCannotWithdrawFully();
>         withdrawalEvent.watch(function(error,log){
> ```

maybe it's just me, but it took me over a minute to understand why we expect this event to occur.
Perhaps in cases like these, where the code isn't obviously "telling the story" it's worthwhile to add a comment *why* we expect the results we expect (here - generating an event)

---

*[test/withdrawUnitTest.js, line 116 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPK5mXXUk2d9O_Xg_M:-KZPK5mXXUk2d9O_Xg_N:bfwrp0j) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L116)):*
> ```JavaScript
>         });
> 
>         yield rosca.withdraw({from: accounts[2]});
> ```

I think there is a wrong comment in ROSCA.sol re this (https://github.com/WeTrustPlatform/rosca-contracts/blob/develop/contracts/ROSCA.sol#L216). When we wrote the code, we thought this is an indication of a bug, but like your test shows, it can happen in a regular course of a ROSCA.
Could you add a fix to the comment in ROSCA.sol?

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPJVlW9JIH3CsUqxIY:-KZPJVlW9JIH3CsUqxIZ:bsyv3ad) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*
> ```JavaScript
>         yield rosca.withdraw({from: accounts[2]});
>         var credit_after = yield rosca.members.call(accounts[2]);
>         return assert.equal(credit_after[0],credit_before[0] - withdrewAmount, "partial withdraw didn't work properly");
> ```

- space after comma
- no need to use ```return```

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKP3zhfdFudMZmDoN:-KZPKP3zhfdFudMZmDoO:b-skielz) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*
> ```JavaScript
>         yield rosca.withdraw({from: accounts[2]});
>         var credit_after = yield rosca.members.call(accounts[2]);
>         return assert.equal(credit_after[0],credit_before[0] - withdrewAmount, "partial withdraw didn't work properly");
> ```

Perhaps also assert the the contract balance is 0 now and that accounts[2] got the money?

---

*[test/withdrawUnitTest.js, line 121 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPOnxnLZf0YNlv93q5:-KZPOnxnLZf0YNlv93q6:b-ng4br1) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L121)):*
> ```JavaScript
>     }));
> 
>     it("calling withdraw when amountToWithdraw < contract.balance", co(function *() {
> ```

also here, let's have the test name reflect the behavior instead of internal vars

---

*[test/withdrawUnitTest.js, line 156 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPP-nyZS7W_YWVv18v:-KZPP-nyZS7W_YWVv18w:b-s7cx6i) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L156)):*
> ```JavaScript
>         var currentRound = yield rosca.currentRound.call();
> 
>         return assert.equal(credit_after[0], currentRound * CONTRIBUTION_SIZE, "withdraw doesn't send the right amount");
> ```

- do we need the return?
- let's also check the balance of the contract and of the recipient?

---

*[test/withdrawUnitTest.js, line 223 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPPXjl1xQVKuN3VsSA:-KZPPXjmAFNiRoIP_1H1:b-9sh92u) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L223)):*
> ```JavaScript
>             rosca.contribute({from: accounts[2], value: CONTRIBUTION_SIZE}),
>             rosca.contribute({from: accounts[1], value: CONTRIBUTION_SIZE}),
>             rosca.contribute({from: accounts[3], value: CONTRIBUTION_SIZE * 3}), // to make sure contract's balance is less than winning bid
> ```

* is the comment a leftover?
* i'm not sure i understand - why 3 times the contribution?

---

*[test/withdrawUnitTest.js, line 237 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPQ1SS7q_bfADDQ-Al:-KZPQ1ST7Jip-9Qc6NIs:b-l756b) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L237)):*
> ```JavaScript
> 
>         yield rosca.withdraw({from: accounts[2]});
>         var credit_after = yield rosca.members.call(accounts[2]);
> ```

this isn't really credit_after, i think it's user. Maybe a clearer approach could be:
var credit_after = yield rosca.members.call(accounts[2])[0];
so that we don't use the interim var

---

*[test/withdrawUnitTest.js, line 241 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPPj5dYh3QZgM4hXv_:-KZPPj5dYh3QZgM4hXva:b-h7bomt) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L241)):*
> ```JavaScript
>         var totalDiscount = yield rosca.totalDiscounts.call();
> 
>         return assert.equal(credit_after[0], (currentRound * CONTRIBUTION_SIZE) - (totalDiscount / MEMBER_COUNT), "withdraw doesn't send the right amount");
> ```

- this test may still pass if totalDiscounts is calculated incorrectly in the contract. I suggest the test computes by itself how much the expected credit is.
- is return needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-:-KZPQFvPaPkHfV4qRwgl:b-92m6e5)*
<!-- Sent from Reviewable.io -->

shine2lay:
Review status: 1 of 5 files reviewed at latest revision, 44 unresolved discussions, some commit checks failed.

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGWyI_PYjlDy0K369:-KZTGWv3uF1dRFPU2qHr:bwe592j) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*

do we really need the "return" here?
</blockquote></details>

Yes, otherwise race condition will happen since its a promise object.

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGpV_GwI3kRNKZBd5:-KZTHsMQ0jXWIvLaOvsF:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*

I suggest that we also run a valid contribute() in the same test (e.g. from accounts[1]) to make sure we're catching the case that a non-member is contributing, and we don't get a throw because of another reason (e.g. if the production code later changes not to allow contribute() before ROSCA starts)
</blockquote></details>

Done.

---

*[test/contributeUnitTest.js, line 26 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPH4EvqZu6lgE8ogNI:-KZTH_p1VbcKqB4Zf6ji:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L26)):*

consider
it("generates a LogContributionMade event after a successful event")
</blockquote></details>

Done.

---

*[test/contributeUnitTest.js, line 37 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPHLd4pWj2PcJ6hsOv:-KZTI8xZxWoFa75UDZo_:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L37)):*

nit: plz put the actual contribution (CONTRIBUTION_SIZE * 0.1) in a constant, so that it's easy to understand why we're expecting a certain amount.
</blockquote></details>

Done.

---

*[test/contributeUnitTest.js, line 54 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPHqi_F65ksVEZQYVO:-KZTNSdRGUplbfp6FeVF:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L54)):*

I suggest contributing twice, and seeing that the credit is the sum of both contributions.
Otherwise, the following will also pass this test:
```members[msg.sender].credit = msg.value;```
( = instead of += )

</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 3 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPI9nyuiDOeTZxTxcw:-KZTNYbXc9gz9R9Yj62r:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L3)):*

nit: would be nice to consistenly add a blank line above contract()
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 6 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIFgCnxHgkTfscTAU:-KZTN_RKp5I2VbDNqZ0Q:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L6)):*

nit: in order to normalize the code, consider 
const MEMBER_COUNT = MEMBER_LIST.length + 1;

(make it less likely for both these vars to fall out of sync in future)
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIX8klzwNnRIIB0-H:-KZTP-4v9j11FKhGLKWQ:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L21)):*

I suggest to first contribute from all members, so it's clear that even in that case withdraw() still throws
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 22 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPEXZuW2qzOyUe3-X2-r2-22:-KZTOTMzzQupd2q95N-y:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L22)):*

Please add the lines to remove this warning
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 36 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIentQPV4yhzzGm5x:-KZTPLtusussfyAt9wch:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L36)):*

same as above, plz put the exact contribution size in a constant, so the test is easier to read
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 80 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKWxqZmDkqeHucDvy:-KZTPT3lOsmmQ1w6SIii:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L80)):*

amountToWithdraw is an implementation detail. Perhaps we can have the test name reflect the behavior, e.g.
it("generates a LogCannotWithdrawFully when the contract balance is less than what the user is entitled to")
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 111 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPJqrOZYblmfbGDr1o:-KZTQHhjkjkuDGb9RAen:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L111)):*

maybe it's just me, but it took me over a minute to understand why we expect this event to occur.
Perhaps in cases like these, where the code isn't obviously "telling the story" it's worthwhile to add a comment *why* we expect the results we expect (here - generating an event)
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPJVlW9JIH3CsUqxIY:-KZTQgLYTUizbLWSOGqu:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*

- space after comma
- no need to use ```return```
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKP3zhfdFudMZmDoN:-KZTYQVKDbXl1VoqzOeJ:buama2i) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*

Perhaps also assert the the contract balance is 0 now and that accounts[2] got the money?
</blockquote></details>

i'll look into the recipient part, it'll be tricky 

---

*[test/withdrawUnitTest.js, line 121 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPOnxnLZf0YNlv93q5:-KZTR7VWmWodgqRwwpmk:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L121)):*

also here, let's have the test name reflect the behavior instead of internal vars
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 156 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPP-nyZS7W_YWVv18v:-KZTR9n5TmiqloXkn1Ks:bi1dyga) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L156)):*

- do we need the return?
- let's also check the balance of the contract and of the recipient?
</blockquote></details>

return removed, checking the recipient's balance is very tricky , since gas cost needs to accounted , i tried to implement it before with no luck, i'll take another look since i should have more knowledge of the system by now 

---

*[test/withdrawUnitTest.js, line 223 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPPXjl1xQVKuN3VsSA:-KZTReGm9os1R_Wkdzy4:b-kihab3) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L223)):*

* is the comment a leftover?
* i'm not sure i understand - why 3 times the contribution?
</blockquote></details>

it was a leftover from when i copy and pasted. 

---

*[test/withdrawUnitTest.js, line 237 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPQ1SS7q_bfADDQ-Al:-KZTVTRvk_6iMMuMXGVk:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L237)):*

this isn't really credit_after, i think it's user. Maybe a clearer approach could be:
var credit_after = yield rosca.members.call(accounts[2])[0];
so that we don't use the interim var
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 241 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPPj5dYh3QZgM4hXv_:-KZTVU4siuMc043Nltum:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L241)):*

- this test may still pass if totalDiscounts is calculated incorrectly in the contract. I suggest the test computes by itself how much the expected credit is.
- is return needed?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28)*

rmerom:
Review status: 1 of 5 files reviewed at latest revision, 44 unresolved discussions, some commit checks failed.

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGWyI_PYjlDy0K369:-KZU1oawdh77vPSWWTwz:b-qzae39) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

Yes, otherwise race condition will happen since its a promise object.
</blockquote></details>

I'll probably need you to explain this to me f2f. I don't  seem to understand what's special about this and why we can't yield it like other statements.

---

*[test/contributeUnitTest.js, line 3 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU2VL0RFjOIzxkVTiA:-KZU2VL0RFjOIzxkVTiB:b4pufn) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/contributeUnitTest.js#L3)):*
> ```JavaScript
> let Promise = require("bluebird");
> let co = require("co").wrap;
> let assert = require("chai").assert;
> ```

yes, i think this makes more sense then what i've done... 

---

*[test/contributeUnitTest.js, line 19 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU22i_laVUHrVkG7_H:-KZU22i_laVUHrVkG7_I:b-uc0o4r) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/contributeUnitTest.js#L19)):*
> ```JavaScript
> 
>         var rosca = yield ROSCATest.new(ROUND_PERIOD_IN_DAYS, CONTRIBUTION_SIZE, DayFromNow, MEMBER_LIST, SERVICE_FEE);
>         yield rosca.contribute({from: accounts[2], value: CONTRIBUTION_SIZE});
> ```

Imagine coming back to this code 6 months from now, breaking your head why one statement would throw and the other won't... consider adding a comment explaining what's going on here.

---

*[test/contributeUnitTest.js, line 27 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU2JDR-wuCJKz78e1P:-KZU2JDR-wuCJKz78e1Q:b-9dn6j) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/contributeUnitTest.js#L27)):*
> ```JavaScript
>     }));
> 
>     it("generates a LogContributionMade event after a successful event", co(function *() {
> ```

my wrong: s/event/contribution

---

*[test/withdrawUnitTest.js, line 6 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIFgCnxHgkTfscTAU:-KZU2xLDkQu6dTNM5Cwy:b-k9mgf8) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L6)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I think you've done it in the other flie, but here it's still missing..

---

*[test/withdrawUnitTest.js, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIX8klzwNnRIIB0-H:-KZU3AHRZMuysDPupF-u:bv7td89) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L21)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thanks, can we "prove" that the throw is caused by the sender being a non-member (and not because of the stage of the ROSCA, say), by issuing a successful withdraw in the same test?

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKP3zhfdFudMZmDoN:-KZUWTLA4jjgKoba004U:bauizwp) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

i'll look into the recipient part, it'll be tricky 
</blockquote></details>

Perhaps we can use web3.eth.getBalance(accounts[2]) in conjunction with assert.closeTo ?

---

*[test/withdrawUnitTest.js, line 156 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPP-nyZS7W_YWVv18v:-KZUX6IzKosSDA0F9Wn2:bb6u032) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L156)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

return removed, checking the recipient's balance is very tricky , since gas cost needs to accounted , i tried to implement it before with no luck, i'll take another look since i should have more knowledge of the system by now 
</blockquote></details>

Cool

---

*[test/withdrawUnitTest.js, line 23 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU33vS_dv8vjTCHyMt:-KZU33vS_dv8vjTCHyMu:b-hw04h6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L23)):*
> ```JavaScript
>         var rosca = yield ROSCATest.new(ROUND_PERIOD_IN_DAYS, CONTRIBUTION_SIZE, DayFromNow, MEMBER_LIST, SERVICE_FEE);
> 
>         Promise.all([
> ```

don't we need to yield this?

---

*[test/withdrawUnitTest.js, line 115 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUW-YEqCYLXprBv2Tf:-KZUW-YEqCYLXprBv2Tg:b-alvkt4) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L115)):*
> ```JavaScript
>             id: new Date().getTime()
>         });
>         rosca.startRound(); // 2nd Member will be able to withdraw POT which is greater than contributionSize
> ```

don't we mean here that the 2nd members is *entitled* to DEFAULT_POT (they actually won't be able to withdraw it at this stage).
Also s/contributionSize/CONTRIBUTION_SIZE

---

*[test/withdrawUnitTest.js, line 130 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUWkuCsWcCdJre0DVE:-KZUWkuD7JWSTNL6gEn8:b-f5gzyp) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L130)):*
> ```JavaScript
>     }));
> 
>     it("generates a LogCannotWithdrawFully when the contract balance is more than what the user is entitled to", co(function *() {
> ```

I think we're expecting this NOT to generate an event in this case?

---

*[test/withdrawUnitTest.js, line 214 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXHJAPiRrg91POW-q:-KZUXHJAPiRrg91POW-r:b-z6luok) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L214)):*
> ```JavaScript
>     }));
> 
>     it("generates a LogCannotWithdrawFully when the contract balance is more than what the user is entitled to while totalDiscount != 0", co(function *() {
> ```

here again i think it should NOT generate the event
(should we actually test the event is not generated?)

---

*[test/withdrawUnitTest.js, line 247 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXlfo15fPM783mLvH:-KZUXlfo15fPM783mLvI:b-oemt55) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L247)):*
> ```JavaScript
> 
>         yield rosca.withdraw({from: accounts[2]});
>         var credit_after = yield rosca.members.call(accounts[2])[0];
> ```

this should be creditAfter. Same for other vars in other tests where we use snake_notation instead of camelCase.

---

*[test/withdrawUnitTest.js, line 251 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXkTmO5uIInZWaZB1:-KZUXkTmO5uIInZWaZB2:b24u103) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L251)):*
> ```JavaScript
>         var totalDiscount = DEFAULT_POT - BID_TO_PLACE;
> 
>         return assert.equal(credit_after, (currentRound * CONTRIBUTION_SIZE) - (totalDiscount / MEMBER_COUNT), "withdraw doesn't send the right amount");
> ```

nit: consider putting (currentRound * CONTRIBUTION_SIZE) - (totalDiscount / MEMBER_COUNT) into a variable called expectedCreditAfter to make it easier to understand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28)*

shine2lay:
Review status: 1 of 5 files reviewed at latest revision, 43 unresolved discussions, some commit checks failed.

---

*[test/contributeUnitTest.js, line 19 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPGWyI_PYjlDy0K369:-KZU_hrIEqINsutOXZi4:bs19ck8) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/contributeUnitTest.js#L19)):*

I'll probably need you to explain this to me f2f. I don't  seem to understand what's special about this and why we can't yield it like other statements.
</blockquote></details>

you can probably yield it but we'll have to change logic to check after yield, instead of catch. Do you have a problem with return? 
Since you wrote a utils function to check for Throws, we can change it in the next PR 

---

*[test/contributeUnitTest.js, line 19 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU22i_laVUHrVkG7_H:-KZUeTsYU3we74l2dFOw:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/contributeUnitTest.js#L19)):*

Imagine coming back to this code 6 months from now, breaking your head why one statement would throw and the other won't... consider adding a comment explaining what's going on here.
</blockquote></details>

Done.

---

*[test/contributeUnitTest.js, line 27 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU2JDR-wuCJKz78e1P:-KZUgWCaKVqPifAjyiJF:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/contributeUnitTest.js#L27)):*

my wrong: s/event/contribution
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 6 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIFgCnxHgkTfscTAU:-KZUgdBg1nKvYOMouyb9:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L6)):*

I think you've done it in the other flie, but here it's still missing..
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPIX8klzwNnRIIB0-H:-KZUh3p3HlWHxeu2eVh-:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L21)):*

Thanks, can we "prove" that the throw is caused by the sender being a non-member (and not because of the stage of the ROSCA, say), by issuing a successful withdraw in the same test?
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 118 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZPKP3zhfdFudMZmDoN:-KZUcHUMGMEFf7D61aCK:bf80n74) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/86b283442ad1efcca13ed9b3bad9b0789b30f6aa/test/withdrawUnitTest.js#L118)):*

Perhaps we can use web3.eth.getBalance(accounts[2]) in conjunction with assert.closeTo ?
</blockquote></details>

well, we'll have to assume the gas price in this case but it should work. I could think of a case where withdraw amount is lower than gas usage, then the test would fail. Unlikely but possible in case where contract balance is very low 

---

*[test/withdrawUnitTest.js, line 23 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZU33vS_dv8vjTCHyMt:-KZUgfDhXmDUAWCeH7jM:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L23)):*

don't we need to yield this?
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 115 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUW-YEqCYLXprBv2Tf:-KZUhGFGrqLSf_OiLMVr:bd8f1yv) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L115)):*

don't we mean here that the 2nd members is *entitled* to DEFAULT_POT (they actually won't be able to withdraw it at this stage).
Also s/contributionSize/CONTRIBUTION_SIZE
</blockquote></details>

after startRound is called, yes they can actually withdraw it. i can change it to entitled if it makes more sense to you

---

*[test/withdrawUnitTest.js, line 130 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUWkuCsWcCdJre0DVE:-KZUiNllKgM_nmKbTZAu:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L130)):*

I think we're expecting this NOT to generate an event in this case?
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 214 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXHJAPiRrg91POW-q:-KZUiTb5yGkqV7DtqWTJ:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L214)):*

here again i think it should NOT generate the event
(should we actually test the event is not generated?)
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 247 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXlfo15fPM783mLvH:-KZUiejzNnzyht68Re7x:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L247)):*

this should be creditAfter. Same for other vars in other tests where we use snake_notation instead of camelCase.
</blockquote></details>

Done.

---

*[test/withdrawUnitTest.js, line 251 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28#-KZUXkTmO5uIInZWaZB1:-KZUif3GG9MeY0-ZigcO:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/e6ca98083e9f2048742ebf42b90f88309f42c497/test/withdrawUnitTest.js#L251)):*

nit: consider putting (currentRound * CONTRIBUTION_SIZE) - (totalDiscount / MEMBER_COUNT) into a variable called expectedCreditAfter to make it easier to understand.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/28)*

rmerom (approved):

PR #46, opened 2016-12-26
Require minimum gap between bids (currently at 98% of previous bid)

rmerom:
Continuing our conversation from the past, maybe percent does make more sense than an absolute gap? 
The pro is that it works better with different pot-sized ROSCAs.
The con we mentioned is that it might produce "ugly" maximum bids, but we can the frontend round those numbers to something more meaningful. 
The 2% required gap sounded to make like something sensible - because one can think of this as a difference in weekly/monthly interest, these numbers seem reasonable.

Please do let me know though if you think we should do something different.

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/46)


shine2lay:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/46#-:-KZwF-_52UjeaoL8l-b8:bnfp4nl)*

shine2lay (approved):

PR #78, opened 2017-01-05
Provide API to the frontend; move many deps to dev

rmerom:
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/78)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 let fs = require('fs');
 let solc = require('solc');
+let path = require('path');
 
 let contractCache;
 
@@ -9,7 +10,7 @@ let WeTrustContract = {
    * Throws if there was an error.
    */
   init: function(done) {
-    let roscaContractLocation = 'contracts/ROSCA.sol';
+    let roscaContractLocation = path.join(__dirname, 'contracts/ROSCA.sol');
 
     fs.readFile(roscaContractLocation, 'utf8', function(err, contractCode) {
       if (err) {
--- a/index.js
+++ b/index.js
@@ -1,26 +1,29 @@
-var fs = require('fs');
+let fs = require('fs');
 let solc = require('solc');
 
 let contractCache;
-let contractErr;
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -0,0 +1,47 @@
...
+let path = require('path');
+
+let contractCache;
+
+let WeTrustContract = {
+  /**
+   * Call this function upon server start. It reads and compiles the contract, and saves it in a cache variable.
+   * Throws if there was an error.
+   */
+  init: function(done) {
```
anzheng67, line 12: I would suggest move this init() and execute it outside of WeTrustContract scope.  Basically, WeTrustContract would only have getContract() for now.

anzheng67: It would be easier for client to use. For example, as a client, I just need to:
let rosca = require('wetrust-rosca-contract');
let contract = rosca.getContract();

Client does not need to handle async callback

index.js
```diff
@@ -0,0 +1,47 @@
...
+    let roscaContractLocation = path.join(__dirname, 'contracts/ROSCA.sol');
+
+    fs.readFile(roscaContractLocation, 'utf8', function(err, contractCode) {
```
anzheng67, line 15: If you prefer to have init() method inside WeTrustContract scope, I would suggest use readFileSync so that client does not have to nest their code in callback when using this library,

index.js
rmerom: Just to understand better: you mean moving the logic of init to rosca-mvp? If so, why do we need getContract all? It only returns a var.

anzheng67: The init() is still in this index.js.  We call it before export so that client does not have to make extra call.  Basically, something like this:

let init = function(){...};
init();
let WeTrustRosca = {...};
exports = WeTrustRosca

anzheng67 (approved):
I am going to merge this implementation and update npm for now.  We will enhance it later.

PR #109, opened 2017-01-22
make random selection fairer

rmerom:
Following Gun's comments, make sure we select unpaid participants fairly when there is no bid:
We select the winner among [x[0]..x[numMembers - currentRound]].

Whenever we find a winner i, we swap x[i] with x[numMembers - currentRound] so that we won't choose them again.

In theory, we can stop checking if a candidate has already been paid because of the above, but I didn't put this in this PR as I wanted to make sure everything works first.

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/109)


shine2lay:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/109#-:-Kb8VRfPNHQeM9X_M7Zh:bnfp4nl)*

shine2lay:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/109)*

shine2lay (approved):

shine2lay:
Have you thought about having a temporary array to hold unpaid members and then use semi-random to choose from that temporary array? 

That was my intuitive solution after hearing Gun's suggestion, i was just wondering what your opinion on that is.

rmerom:
Yes, I think that's also a suggestion Gun made. I did the swapping because it has 2 benefits: one is that it requires less space, and the other is that all the unpaid members are now stored in the beginning of the array, which may be helpful when we add future features.

PR #111, opened 2017-01-27
fix getParticipantBalance to show proper values.

shine2lay:
getParticipantBalance doesn't show proper values for users in "debt" after the endofROSCA

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/111)

### DIFF
--- a/contracts/ROSCA.sol
+++ b/contracts/ROSCA.sol
@@ -435,7 +435,7 @@ contract ROSCA {
   function getParticipantBalance(address user) onlyFromMember external constant returns(int256) {
     int256 totalCredit = int256(members[user].credit + totalDiscounts);
 
-    if (members[user].debt) {
+    if (members[user].debt && !endOfROSCA) {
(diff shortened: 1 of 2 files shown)


shine2lay:
please publish a new version of contract after merging

rmerom (commented):
Good that you caught this edge case!
At end of ROSCA, shouldn't a delinquent participant be able to withdraw only his contributions, not his entire credit?

Also would be great, though not completely required, to add a test to getParticipantBalance that covers this case.

shine2lay:
well, this fix wouldn't allow user to withdraw his entire credit.
we can ignore this line because, at the endofROSCA, 

``` totalDebit = int256(currentRound * contributionSize); ```
would be the same as Default pot, the return value should be contributions

rmerom:
You're right!
As this is not completely intuitive (the fact that the winning will zero out the debit), can we add a comment to that effect?

rmerom (approved):

PR #120, opened 2017-02-09
Up ver to 0.0.14

rmerom:
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/120)


rmerom:
@morelazers you should be able to use 0.0.14 for the new fee mechanism

PR #129, opened 2017-03-05
First stab at ERC20 tokenable ROSCA - do not use this contract yet

rmerom:
This PR's goal is to see if we're all on the same page as to how to integrate ERC20 tokens into the ROSCA and make sure we use the standard correctly.
A later PR will add tests, so do not use this contract as of right now.

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/129)


shine2lay:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/tokenableROSCA.sol, line 348 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_9wqvEkekLoZhkAMu:-Ke_9wqwkfXgjdX4FSi2:b7z444t) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L348)):*
> ```
>       throw;
>     }
>     uint256 value = (isEthRosca ? msg.value : tokenContract.allowance(msg.sender, address(this)));
> ```

what happens if tokenContract.allowance() is not a function (i.e tokenContract is not a valid ERC20 token)?

---

*[contracts/tokenableROSCA.sol, line 352 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_AULdScOjCZkAuC8e:-Ke_AULdScOjCZkAuC8f:b28s1b) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L352)):*
> ```
>       throw;
>     }
>     if (isEthRosca) {
> ```

it maybe possible to consolidate this with above , line 348

---

*[contracts/tokenableROSCA.sol, line 421 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BRda6cxsZiGcXOQm:-Ke_BRda6cxsZiGcXOQn:b-sx3nz6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L421)):*
> ```
> 
>   // Sends funds (ETH or tokens) to the specified recipient. Returns whether successful.
>   function sendFundsTo(address recipient, uint256 value) internal returns (bool) {
> ```

would recipient always be "msg.sender"?

---

*[contracts/tokenableROSCA.sol, line 492 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BvFp8SW8nGxGu6zs:-Ke_BvFp8SW8nGxGu6zt:bf7lrwj) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L492)):*
> ```
>    */
>   function getBalance() internal constant returns (uint256) {
>     bool isEthRosca = (tokenContract == address(0));
> ```

have you thought about making this global variable? i know it might be more expensive to have a global variable but since we only have to set isEthRosca once, it might be cheaper in the long run.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129)*

rmerom:
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/tokenableROSCA.sol, line 348 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_9wqvEkekLoZhkAMu:-Ke_ixov0g_o-CTM8iw2:b-opru79) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L348)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

what happens if tokenContract.allowance() is not a function (i.e tokenContract is not a valid ERC20 token)?
</blockquote></details>

As Line 177 mentions, it is the responsibility of the creator to check that the contract is ERC20 compliant and safe.
On a practical side, in our MVP, the FE can limit the tokens allowed.
Does that answer the question?

---

*[contracts/tokenableROSCA.sol, line 352 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_AULdScOjCZkAuC8e:-Ke_tAPTdMpuxqlKKyX3:b-enbf63) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L352)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

it maybe possible to consolidate this with above , line 348
</blockquote></details>

I'm not sure how you mean. Can you give a code example with a brief explanation of how it's better?

---

*[contracts/tokenableROSCA.sol, line 421 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BRda6cxsZiGcXOQm:-Ke_tGHV6YIBFnjICRdC:blpbrnv) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L421)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

would recipient always be "msg.sender"?
</blockquote></details>

good point! changed.

---

*[contracts/tokenableROSCA.sol, line 492 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BvFp8SW8nGxGu6zs:-Ke_tkXtn4BcoMbmQGJn:bttzu81) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L492)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

have you thought about making this global variable? i know it might be more expensive to have a global variable but since we only have to set isEthRosca once, it might be cheaper in the long run.
</blockquote></details>

yeah, each option has its own pros and cons (much lower one-time deployment costs vs. slight less lower many-time maintenance cost).
If you feel strongly that adding a global variable is better, I can change it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129)*

shine2lay:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/tokenableROSCA.sol, line 348 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_9wqvEkekLoZhkAMu:-Ke_xY-dVeA9idxfdImp:b-tqr3uz) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L348)):*

As Line 177 mentions, it is the responsibility of the creator to check that the contract is ERC20 compliant and safe.
On a practical side, in our MVP, the FE can limit the tokens allowed.
Does that answer the question?
</blockquote></details>

i guess my question is "what is the default behavior when calling a non-existent function?"

---

*[contracts/tokenableROSCA.sol, line 352 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_AULdScOjCZkAuC8e:-Ke_wuJxlawNeGFNNFDn:b-ic33o6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L352)):*

I'm not sure how you mean. Can you give a code example with a brief explanation of how it's better?
</blockquote></details>

uint 256 value; 

if (isEthRosca) {
  return msg.value
} else {
  value = tokenContract.allowance(address(this))
}

i mentioned this because there were 2 places where "isEthRosca" is being checked

---

*[contracts/tokenableROSCA.sol, line 492 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BvFp8SW8nGxGu6zs:-Ke_yMzv_0QEtOsNMteZ:bnhb6nh) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L492)):*

yeah, each option has its own pros and cons (much lower one-time deployment costs vs. slight less lower many-time maintenance cost).
If you feel strongly that adding a global variable is better, I can change it.
</blockquote></details>

i think global variable in this case might be better since we only need to set it once in the constructor and never change it again

---

*[contracts/tokenableROSCA.sol, line 349 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_yjYyG_x7r_2Dqi9S:-Ke_yjYyG_x7r_2Dqi9T:b-vozonw) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/de26d2c41e5157defcf622776ca421e566833d45/contracts/tokenableROSCA.sol#L349)):*
> ```
>     }
>     uint256 value = (isEthRosca ? msg.value : tokenContract.allowance(msg.sender, address(this)));
>     if (value <= 0) {
> ```

i don't think this will ever be negative since ```value``` is uint256

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129)*

rmerom:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/tokenableROSCA.sol, line 348 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_9wqvEkekLoZhkAMu:-KeaJlh9S9ry2Lax0uSO:bpfik8v) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L348)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

i guess my question is "what is the default behavior when calling a non-existent function?"
</blockquote></details>

http://ethereum.stackexchange.com/questions/1416/what-happens-if-you-call-a-non-existent-function-of-another-contract-that-has-no

---

*[contracts/tokenableROSCA.sol, line 352 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_AULdScOjCZkAuC8e:-KeaK5j7GGb4fRJ38893:boziu9w) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L352)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

uint 256 value; 

if (isEthRosca) {
  return msg.value
} else {
  value = tokenContract.allowance(address(this))
}

i mentioned this because there were 2 places where "isEthRosca" is being checked
</blockquote></details>

right, but i think we still want to check if `value == 0` which makes the required code about as complex as the current one.

---

*[contracts/tokenableROSCA.sol, line 492 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_BvFp8SW8nGxGu6zs:-KeaKQtyseP6KgK-FhVC:bdjxwdb) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/f4f55c976aca4fc2dfa8968034cd9c2a67343648/contracts/tokenableROSCA.sol#L492)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

i think global variable in this case might be better since we only need to set it once in the constructor and never change it again
</blockquote></details>

it'll probably happen many times that one of us does something in a certain way and the other thinks he would have done it in a different way, and there are pros and cons to both ways, but nothing significant.
Usually I prefer to leave it up to the PR author to choose in order to reduce friction, unless I feel strongly that my way would be better.
So as I said, if you feel strongly that a global variable is better (not just prefer it), plz let me know and I'll update.

---

*[contracts/tokenableROSCA.sol, line 349 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129#-Ke_yjYyG_x7r_2Dqi9S:-KeaK1GfY1EYWDtTEs6X:b-g7lpia) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/de26d2c41e5157defcf622776ca421e566833d45/contracts/tokenableROSCA.sol#L349)):*
<details><summary><i>Previously, shine2lay wrote…</i></summary><blockquote>

i don't think this will ever be negative since ```value``` is uint256
</blockquote></details>

Good point. Changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/129)*

shine2lay (approved):

contracts/tokenableROSCA.sol
```diff
@@ -330,6 +338,26 @@ contract ROSCA {
     return amount * (1000 - serviceFeeInThousandths) / 1000;
   }
 
+  // Validates a non-zero contribution from msg.sender and returns
+  // the amount.
+  function validateAndReturnContribution() internal returns (uint256) {
+    bool isEthRosca = (tokenContract == address(0));
+    if (!isEthRosca && msg.value > 0) {  // token ROSCAs should not accept ETH
+      throw;
```
morelazers, line 346: Does throw refund the ETH/Tokens?

contracts/tokenableROSCA.sol
```diff
@@ -376,18 +405,27 @@ contract ROSCA {
...
     }
-    lowestBid = bidInWei;
+    lowestBid = bid;
     winnerAddress = msg.sender;
     LogNewLowestBid(lowestBid, winnerAddress);
   }
 
+  // Sends funds (either ETH or tokens) to msg.sender. Returns whether successful.
+  function sendFundsToMsgSender(uint256 value) internal returns (bool) {
+    bool isEthRosca = (tokenContract == address(0));
```
morelazers, line 422: iirc I think you can use modifiers to instantiate the `isEthRosca` bool, just a thought since this check is in a couple of places.

But I do think that using a modifier for something this this reduces readbility, I like the current approach better

morelazers (commented):
This looks awesome! I didn't read through the whole contract only the changes, but I don't see any problems.

contracts/tokenableROSCA.sol
rmerom: yes, it does

contracts/tokenableROSCA.sol
rmerom: yeah i agree with these considerations

PR #130, opened 2017-03-06
First attempt at Trustcoin

morelazers:
I don't have a dev environment quite set up yet, so it's not tested or anything, just a quick mock up of the various functionalities to see if I have missed/overlooked anything

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/130)


contracts/lib/SafeMath.sol
```diff
@@ -0,0 +1,53 @@
+pragma solidity ^0.4.8;
```
rmerom: You might want to add attribution to where we got this from (assuming you didn't write it from scratch)

contracts/Trustcoin.sol
```diff
@@ -0,0 +1,112 @@
...
+  string public symbol = 'TRST';
+  string public version = 'TRST1.0';
+  uint256 public totalSupply = 100000000; // One hundred million
+  uint256 public totalMigrated;
+  address public newToken;
+
+  mapping(address => uint) balances;
+  mapping (address => mapping (address => uint)) allowed;
+
+  bool migrationActive;
```
rmerom: for extra security, please explicitly state for each variable and function whether they're internal, public or external.

contracts/Trustcoin.sol
```diff
@@ -0,0 +1,112 @@
...
+
+  //
+  // Migration methods
+  //
+
+  /**
+   *  Sets the owner for the migration behaviour
+   *  @param {address} _master Address of the migration controller
+   */
+  function setMigrationMaster(address _master) external {
```
rmerom: I would name this `changeMigrationMaster` as the MigrationMaster is set for the first time upon creation.

contracts/Trustcoin.sol
```diff
@@ -0,0 +1,112 @@
...
+    if (migrationActive) throw;
+    migrationActive = true;
+  }
+
+  /**
+   *  Sets the address of the new token contract, so we know who to
+   *  accept migrate() calls from
+   *  @param {address} _newToken Address of the new Trustcoin contract
+   */
+  function setNewToken(address _newToken) external {
```
rmerom: can we merge this with `allowMigrations` or are these 2 different steps?

contracts/Trustcoin.sol
```diff
@@ -0,0 +1,112 @@
...
+    if (msg.sender != migrationMaster) throw;
```
rmerom: consider using a [modifier](http://solidity.readthedocs.io/en/develop/common-patterns.html#restricting-access) for many of the preconditional checks that appear more than once

contracts/Trustcoin.sol
```diff
@@ -0,0 +1,112 @@
...
+    newToken = _newToken;
+  }
+
+  /**
+   *  Burns the tokens from an address and increments the totalMigrated
+   *  by the same value
+   *  @param {address} _from Address which holds the tokens
+   *  @param {uint256} _value Number of tokens to be migrated
+   */
+  function migrate(address _from, uint256 _value) external {
```
rmerom: since this is called by the new contract, perhaps `discardTokens` or something similar might be a more appropraite name, as this function explicitly does not migrate the tokens?

contracts/ExampleTrustcoin2.sol
```diff
@@ -0,0 +1,117 @@
...
+  bool migrationActive;
+  address migrationMaster;
+
+  event Transfer(address from, address to, uint256 value);
+  event Approval(address from, address to, uint256 value);
+  event Migrate(address owner, uint256 value);
+  event MigrateSuccess(address owner, uint256 value);
+  event MigrationFinalized();
+
+  function Trustcoin2(address _oldToken, address _migrationMaster, bool _allowOldMigrations, uint256 _allowOldMigrationsUntil) {
```
rmerom: Consider having less parameters here and having more hardcoded `constant`s, as this would likely probably not be a general-purpose contract, but rather a one-off. In that case `constant`s can help keep the code shorter and simpler.

rmerom: Specifically
- I find it hard to understand when `allowOldMigrations` might be false.
- Having checks on `allowOldMigrationsUntil` seems like unnecessary code. When we deploy the contract with an "until date", users will be able to examine the whole contract and see if the date is far enough in the future. I can't see how code checks that check a constructor constant for one-off contract could help.

contracts/ExampleTrustcoin2.sol
```diff
@@ -0,0 +1,117 @@
...
+  function setMigrationStatus(bool _active) external {
+    if (msg.sender != migrationMaster) throw;
+    migrationActive = _active;
+  }
+
+  function migrate(uint256 _value) external {
+    if (!migrationActive) throw;
+    if (_value == 0) throw;
+    if (_value > balances[msg.sender]) throw;
+    balances[msg.sender] = safeSub(balances[msg.sender], _value);
```
rmerom: shouldn't this and some of the next lines be `safeAdd`?

contracts/ExampleTrustcoin2.sol
```diff
@@ -0,0 +1,117 @@
...
+    totalMigrated = safeSub(totalMigrated, _value);
+    Migrate(msg.sender, _value);
+  }
+
+  /**
+   *  Migrates the specified token balance from msg.sender in the old contract
+   *  to the new contract
+   *  @param {uint256} _value Number of tokens to be migrated
+   */
+  function migrateOldTokens(uint256 _value) external {
```
rmerom: how is this related to the `migrate` above?

contracts/ExampleTrustcoin2.sol
rmerom: actually, is this whole method redundant?

contracts/ExampleTrustcoin2.sol
```diff
@@ -0,0 +1,117 @@
...
+    if (!allowOldMigrations) throw;
+    if (_value == 0) throw;
+    Trustcoin(oldToken).migrate(msg.sender, _value);
+    totalSupply = safeAdd(totalSupply, _value);
+    balances[msg.sender] = safeAdd(balances[msg.sender], _value);
+    MigrateSuccess(msg.sender, _value);
```
rmerom: Can we call this just plain old `Migrate`? There's no `MigrateFailure`...

rmerom (commented):
Looks like a good start!
I mainly looked only at the API util you give the green light to look at the rest.

contracts/ExampleTrustcoin2.sol
morelazers: This is the method which continues to support migrations to new contracts, it should have matched up with the method in the other file, but I overlooked a couple of details.

contracts/ExampleTrustcoin2.sol
morelazers: This is the method which users call to migrate their old tokens to the new contract. The `migrate()` function above I have renamed `discardTokens()` to be in line with the other file.

Do you think this should be renamed to `migrate()`?

contracts/ExampleTrustcoin2.sol
morelazers: Hmm, it clashes with the other Migrate() event, which is for a separate purpose. I will think about how best to rename them

rmerom:
Review status: 0 of 4 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[contracts/ExampleTrustcoin2.sol, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefJdV2rW6K9AQy0iIc:-KefJdV2rW6K9AQy0iId:b-uqtq8i) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L21)):*
> ```
>   string public symbol = 'TRST2';
>   string public version = 'TRST2.0';
>   uint256 public totalSupply; // Begins at 0, but increments as old tokens are migrated (ERC20)
> ```

suggestion: as old token are migrated *into this contract*
this is because migration happen into, and possibly later, also out of, this contract

---

*[contracts/ExampleTrustcoin2.sol, line 22 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWAAKBYlDFOvANfoU:-KedWAAKBYlDFOvANfoV:bpitlhi) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L22)):*
> ```
>   string public version = 'TRST2.0';
>   uint256 public totalSupply; // Begins at 0, but increments as old tokens are migrated (ERC20)
>   address public oldToken = '0x123f681646d4a755815f9cb19e1acc8565a0c2ac'; // Address of our old Trustcoin token contract (this is just a random address)
> ```

please use `constant` where appropraite

---

*[contracts/ExampleTrustcoin2.sol, line 24 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWFwrmqAAAn76YUyT:-KedWFwsscXZJWaDQ5y1:b-brremg) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L24)):*
> ```
>   address public oldToken = '0x123f681646d4a755815f9cb19e1acc8565a0c2ac'; // Address of our old Trustcoin token contract (this is just a random address)
>   bool public allowOldMigrations = true; // Is set to false when we finalize migration
>   uint256 public allowOldMigrationsUntil = (now + 26 weeks); // Hardcode a timestamp here?
> ```

I think that's good enough - self explanatory and since it is public, anyone can read it.

---

*[contracts/ExampleTrustcoin2.sol, line 29 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefK6hJmbeJA8gMDl1u:-KefK6hKgJIPsspRjKKp:b-65hyo6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L29)):*
> ```
>   mapping (address => mapping (address => uint)) public allowed; // (ERC20)
> 
>   // Variables supporting the migration to a new contract (Trustcoin3)
> ```

should we extract this migration functionality to a base contract, e.g. `upgradable` or `migratable`?

---

*[contracts/ExampleTrustcoin2.sol, line 36 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKKh7o8EO7DMXGOng:-KefKKh7o8EO7DMXGOnh:bw1jf8d) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L36)):*
> ```
> 
>   event Discard(address owner, uint256 value);
>   event Migrate(address owner, uint256 value);
> ```

Since the migration can happen both ways, can we name this more explicitly, e.g. IncomingMigration ?

---

*[contracts/ExampleTrustcoin2.sol, line 50 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKnCpFx2arQCroRdC:-KefKnCq2Fm-jS-x3EDe:bvzb9rz) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L50)):*
> ```
> 
>   /**
>    *  (ERC20)
> ```

for the ERC20 comments, as long as the ERC20 interface has these comments, I would leave only `see ERC20` here, or use something similar to Java's `@override`

This is in order to avoid duplication of the comments

---

*[contracts/ExampleTrustcoin2.sol, line 65 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKcwMDhMHcOk_g2Gn:-KefKcwN4atDQCV8h7j6:bdm0vt2) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L65)):*
> ```
>   /**
>    *  (ERC20)
>    *  Allows contracts to send tokens on our behalf
> ```

is this the right comment for `transferFrom` ?

---

*[contracts/ExampleTrustcoin2.sol, line 132 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefLPW3Ykixnd1tINiH:-KefLPW3Ykixnd1tINiI:b-vz6jp1) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L132)):*
> ```
>    *  @param {address} _newToken Address of the new Trustcoin contract
>    */
>   function setNewToken(address _newToken) onlyFromMigrationMaster external {
> ```

consider s/setNewToken/setNewTokenAddress

---

*[contracts/ExampleTrustcoin2.sol, line 145 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefLiYMUv2NmO3FQPdW:-KefLiYMUv2NmO3FQPdX:b5u214s) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L145)):*
> ```
>    *  @param {uint256} _value Number of tokens to be migrated
>    */
>   function discardTokens(address _from, uint256 _value) external {
> ```

nit: consider moving this down the file, so that users of the contract first see the public API meant for them.

---

*[contracts/Trustcoin.sol, line 112 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefODJq_-Kus-1IBLfu:-KefODJq_-Kus-1IBLfv:bc8w7s6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/Trustcoin.sol#L112)):*
> ```
> 
>   /**
>    *  Sets the owner for the migration behaviour
> ```

s/Sets/Changes

---

*[contracts/Trustcoin.sol, line 133 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefOHUxAYgWxX-Cz725:-KefOHUxAYgWxX-Cz726:b8e30a1) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/Trustcoin.sol#L133)):*
> ```
> 
>   /**
>    *  Burns the tokens from an address and increments the totalMigrated
> ```

might be useful to add
"called by the new contract when tokens are migrated"

(otherwise readers who just look at the API might wonder why such a method is external)

---

*[contracts/lib/ERC20.sol, line 2 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefO_qC_ZI_BSglWGFs:-KefO_qC_ZI_BSglWGFt:b3ug764) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/lib/ERC20.sol#L2)):*
> ```
> pragma solidity ^0.4.8;
> ```

can we reuse
https://github.com/WeTrustPlatform/rosca-contracts/blob/develop/contracts/deps/ERC20TokenInterface.sol ?

---

*[contracts/lib/SafeMath.sol, line 1 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgp-r1-1:-KefOfCwmID2PE0DoN3j:bb7mm0n) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/lib/SafeMath.sol#L1)):*

You might want to add attribution to where we got this from (assuming you didn't write it from scratch)
</blockquote></details>

ping

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

morelazers:
Review status: 0 of 3 files reviewed at latest revision, 22 unresolved discussions.

---

*[contracts/ExampleTrustcoin2.sol, line 30 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgm-r1-30:-KehCcKpM9NJTWJMTsJe:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/ExampleTrustcoin2.sol#L30)):*

Specifically
- I find it hard to understand when `allowOldMigrations` might be false.
- Having checks on `allowOldMigrationsUntil` seems like unnecessary code. When we deploy the contract with an "until date", users will be able to examine the whole contract and see if the date is far enough in the future. I can't see how code checks that check a constructor constant for one-off contract could help.
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 21 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefJdV2rW6K9AQy0iIc:-KehB0hBnGLGJWAbgnqX:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L21)):*

suggestion: as old token are migrated *into this contract*
this is because migration happen into, and possibly later, also out of, this contract

</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 22 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWAAKBYlDFOvANfoU:-KehBPDfBpEH3sWUSXgs:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L22)):*

please use `constant` where appropraite
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 24 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWFwrmqAAAn76YUyT:-KehC-98yNnhwC50lKnC:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L24)):*

I think that's good enough - self explanatory and since it is public, anyone can read it.
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 36 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKKh7o8EO7DMXGOng:-KehCYk1Wg2r4w7bgb8c:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L36)):*

Since the migration can happen both ways, can we name this more explicitly, e.g. IncomingMigration ?
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 50 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKnCpFx2arQCroRdC:-KehBvyLjArQcr9qJpt5:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L50)):*

for the ERC20 comments, as long as the ERC20 interface has these comments, I would leave only `see ERC20` here, or use something similar to Java's `@override`

This is in order to avoid duplication of the comments
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 65 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefKcwMDhMHcOk_g2Gn:-KehCiYXxk9guIO4FQOU:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L65)):*

is this the right comment for `transferFrom` ?
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 132 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefLPW3Ykixnd1tINiH:-KehCoDB8IUs2OFC4dcn:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L132)):*

consider s/setNewToken/setNewTokenAddress
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 145 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefLiYMUv2NmO3FQPdW:-KehCycr3CRC1GPR2sYq:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L145)):*

nit: consider moving this down the file, so that users of the contract first see the public API meant for them.
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 19 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgn-r1-19:-KehDazk-qEemjfnfHsQ:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/Trustcoin.sol#L19)):*

for extra security, please explicitly state for each variable and function whether they're internal, public or external. 
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 69 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgn-r1-69:-KehDm2KFludr4OxcdGI:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/Trustcoin.sol#L69)):*

I would name this `changeMigrationMaster` as the MigrationMaster is set for the first time upon creation.
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 89 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgn-r1-89:-KehDmi7rmfKvDeMC2TX:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/Trustcoin.sol#L89)):*

can we merge this with `allowMigrations` or are these 2 different steps?
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 90 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgn-r1-90:-KehDnAEDPxKViGq3aFD:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/Trustcoin.sol#L90)):*

consider using a [modifier](http://solidity.readthedocs.io/en/develop/common-patterns.html#restricting-access) for many of the preconditional checks that appear more than once
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 101 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgn-r1-101:-KehETPvKlTodUFT0H9K:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/Trustcoin.sol#L101)):*

since this is called by the new contract, perhaps `discardTokens` or something similar might be a more appropraite name, as this function explicitly does not migrate the tokens?
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 112 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefODJq_-Kus-1IBLfu:-KehDk4jYjTPqK_Qg4dy:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/Trustcoin.sol#L112)):*

s/Sets/Changes
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 133 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefOHUxAYgWxX-Cz725:-KehESToCKk5YZUiHXvj:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/Trustcoin.sol#L133)):*

might be useful to add
"called by the new contract when tokens are migrated"

(otherwise readers who just look at the API might wonder why such a method is external)

</blockquote></details>

Done.

---

*[contracts/lib/ERC20.sol, line 2 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefO_qC_ZI_BSglWGFs:-KehFUdacguwc1QFUkd7:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/lib/ERC20.sol#L2)):*

can we reuse
https://github.com/WeTrustPlatform/rosca-contracts/blob/develop/contracts/deps/ERC20TokenInterface.sol ?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

morelazers:
Review status: 0 of 3 files reviewed at latest revision, 20 unresolved discussions, some commit checks failed.

---

*[contracts/ExampleTrustcoin2.sol, line 29 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefK6hJmbeJA8gMDl1u:-KehGYW1NB7Uf28PrRI2:b-gj3c04) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L29)):*

should we extract this migration functionality to a base contract, e.g. `upgradable` or `migratable`?
</blockquote></details>

I thought about doing this (would be awesome to write a standard for ERC20 upgrades), but I'm not 100% sure on how to proceed. As far as I understand, either the base contract would be something similar to a pure function library, returning items which would need to be taken care of in the token contract anyway, or the base contract would have to be the one keeping hold of the state of the token migrations.

If you have some thoughts on how we can do this without drastically reducing readability or overcomplicating I would love to hear them.

---

*[contracts/deps/SafeMath.sol, line 1 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgp-r1-1:-KehHeZ5ULmFe6OoYeUG:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/deps/SafeMath.sol#L1)):*

ping
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

rmerom:
Review status: 0 of 3 files reviewed at latest revision, 19 unresolved discussions.

---

*[contracts/ExampleTrustcoin2.sol, line 85 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgm-r1-85:-KejA9ZO8SFXMW6Ikf82:b-ognr9c) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/ExampleTrustcoin2.sol#L85)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

This is the method which continues to support migrations to new contracts, it should have matched up with the method in the other file, but I overlooked a couple of details.
</blockquote></details>

yes, sorry, i completely misunderstood when i was reading this

---

*[contracts/ExampleTrustcoin2.sol, line 96 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgm-r1-96:-KejANzjudZrLQPyc7AE:bm37tko) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/ExampleTrustcoin2.sol#L96)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

This is the method which users call to migrate their old tokens to the new contract. The `migrate()` function above I have renamed `discardTokens()` to be in line with the other file.

Do you think this should be renamed to `migrate()`?
</blockquote></details>

Nope, it's fine as it is. Same confusion of mine as above

---

*[contracts/ExampleTrustcoin2.sol, line 102 at r1](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Ke_N420RPdHRtn0rkgm-r1-102:-KejAU6uei5VKxozpUc1:bb6ywmx) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d01982f179c420379adc6cd368058a94c8fa3e79/contracts/ExampleTrustcoin2.sol#L102)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

Hmm, it clashes with the other Migrate() event, which is for a separate purpose. I will think about how best to rename them
</blockquote></details>

LGTM

---

*[contracts/ExampleTrustcoin2.sol, line 24 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWFwrmqAAAn76YUyT:-Kej8zbtIKOxLnwFVkzH:bm4vft0) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L24)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Actually, the name is not 100% accurate. This date is when MigrationMaster is *allowed* to end the migration, not necessarily when it ends.

---

*[contracts/ExampleTrustcoin2.sol, line 29 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KefK6hJmbeJA8gMDl1u:-Kej8YgVCfCagNbjlmpl:b-8uw6y8) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L29)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

I thought about doing this (would be awesome to write a standard for ERC20 upgrades), but I'm not 100% sure on how to proceed. As far as I understand, either the base contract would be something similar to a pure function library, returning items which would need to be taken care of in the token contract anyway, or the base contract would have to be the one keeping hold of the state of the token migrations.

If you have some thoughts on how we can do this without drastically reducing readability or overcomplicating I would love to hear them.
</blockquote></details>

Hmmm.. that's a good point. What if we take the second approach, that is that the base contract defined the state variables that have to do with the token migrations. Since these are public, AFAIR the derived contract can also access them.

---

*[contracts/ExampleTrustcoin2.sol, line 23 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Kej8na5LPesstiH9d_y:-Kej8na5LPesstiH9d_z:b-6jyfob) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/ExampleTrustcoin2.sol#L23)):*
> ```
>   uint256 public totalSupply; // Begins at 0, but increments as old tokens are migrated into this contract (ERC20)
>   address public constant oldToken = 0x6651fdb9d5d15ca55cc534ee5fa6c3432acdf15b; // Address of our old Trustcoin token contract (this is just a random address)
>   bool public allowOldMigrations = true; // Is set to false when we finalize migration
> ```

nit: as we're using 'incomingMigrations' in the events, can we use this term consistently across the board?

---

*[contracts/ExampleTrustcoin2.sol, line 103 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejA0Mur_mit-hGjXi6:-KejA0Mv69LTQYOKAaXT:b-xmcgka) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/ExampleTrustcoin2.sol#L103)):*
> ```
>   function setNewTokenAddress(address _newToken) onlyFromMigrationMaster external {
>     if (newToken != 0) throw; // Ensure we haven't already set the new token
>     if (_newToken == 0) throw; // Paramater validation
> ```

nit: comment may be redundant..

---

*[contracts/Trustcoin.sol, line 17 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejCzvNIPbPxtMPpFK6:-KejCzvNIPbPxtMPpFK7:b-ze4q0j) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L17)):*
> ```
> 
>   string public constant name = 'Trustcoin';
>   uint8 public constant decimals = 18;
> ```

maybe add "same as eth" in a comment :)

---

*[contracts/Trustcoin.sol, line 20 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejD82_OoM8xPkZJ9Qe:-KejD82_OoM8xPkZJ9Qf:bbb0gk1) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L20)):*
> ```
>   string public constant symbol = 'TRST';
>   string public constant version = 'TRST1.0';
>   uint256 public totalSupply = 100000000; // One hundred million (ERC20)
> ```

I can't really understand why almost all code i see in ETH uses things like 100000000 instead of 1e8, which I find easier to read and verify. Can you enlighten me? (this is not meant to be cynical, I'm really wondering what I'm missing)

---

*[contracts/Trustcoin.sol, line 21 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDXvy713bZUs95QQV:-KejDXvzvtoak58emBew:b-se5k2l) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L21)):*
> ```
>   string public constant version = 'TRST1.0';
>   uint256 public totalSupply = 100000000; // One hundred million (ERC20)
>   uint256 public totalMigrated; // Begins at 0 and increments if tokens are migrated to a new contract
> ```

s/if/as

---

*[contracts/Trustcoin.sol, line 22 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDZgvk_oMG_xzYE7R:-KejDZgvk_oMG_xzYE7S:b-70tdp9) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L22)):*
> ```
>   uint256 public totalSupply = 100000000; // One hundred million (ERC20)
>   uint256 public totalMigrated; // Begins at 0 and increments if tokens are migrated to a new contract
>   address public newToken; // Address of the new token contract
> ```

consider s/newToken/newTokenAddress

---

*[contracts/Trustcoin.sol, line 24 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDwfPjnz_PTi7q3bx:-KejDwfQJXutwK1ZChGM:brvupf5) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L24)):*
> ```
>   address public newToken; // Address of the new token contract
> 
>   mapping(address => uint) public balances; // (ERC20)
> ```

based on feedback we got, we tried to stick with explicit variable sizes in our other contract (e.g. `uint256` instead of `uint`) to make the var's size stand out more for readability and security. You might want to do that too (not a must)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

morelazers:
Review status: 0 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[contracts/ExampleTrustcoin2.sol, line 24 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWFwrmqAAAn76YUyT:-KejSlH4IN7Zdg7MD9hM:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L24)):*

Actually, the name is not 100% accurate. This date is when MigrationMaster is *allowed* to end the migration, not necessarily when it ends.
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 23 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-Kej8na5LPesstiH9d_y:-KejRuGc_Rrht7FywrOk:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/ExampleTrustcoin2.sol#L23)):*

nit: as we're using 'incomingMigrations' in the events, can we use this term consistently across the board?
</blockquote></details>

Done.

---

*[contracts/ExampleTrustcoin2.sol, line 103 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejA0Mur_mit-hGjXi6:-KejTBbs5glZSHUU5b8e:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/ExampleTrustcoin2.sol#L103)):*

nit: comment may be redundant..
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 17 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejCzvNIPbPxtMPpFK6:-KejTQLNgXKBKtrc9ZZn:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L17)):*

maybe add "same as eth" in a comment :)
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 20 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejD82_OoM8xPkZJ9Qe:-KejTRRJczVe7CaEF4Wf:b-380g6l) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L20)):*

I can't really understand why almost all code i see in ETH uses things like 100000000 instead of 1e8, which I find easier to read and verify. Can you enlighten me? (this is not meant to be cynical, I'm really wondering what I'm missing)
</blockquote></details>

Does Solidity support '1e' notation? I can't find it anywhere in the docs

---

*[contracts/Trustcoin.sol, line 21 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDXvy713bZUs95QQV:-KejTsVWCyyLFyXViDWp:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L21)):*

s/if/as
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 22 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDZgvk_oMG_xzYE7R:-KejU8cfj3-L-sc3RVMy:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L22)):*

consider s/newToken/newTokenAddress
</blockquote></details>

Done.

---

*[contracts/Trustcoin.sol, line 24 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDwfPjnz_PTi7q3bx:-KejURIUYj3QpqPgrkWa:b-896fix) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L24)):*

based on feedback we got, we tried to stick with explicit variable sizes in our other contract (e.g. `uint256` instead of `uint`) to make the var's size stand out more for readability and security. You might want to do that too (not a must)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

rmerom:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[contracts/ExampleTrustcoin2.sol, line 24 at r2](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KedWFwrmqAAAn76YUyT:-Kek3dK_prTykZ8QPeMp:beq439j) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/d6950852b3c35d1143d196b951f64d07f699e84c/contracts/ExampleTrustcoin2.sol#L24)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Now that we're moving the `migrate` function to the old contract, it's probably trivial to note that this var should be in the old contract. The upside of it is that it gives contributors more confidence because we commit to the 6 month period right now, not when we provide the new contract.

---

*[contracts/Trustcoin.sol, line 20 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejD82_OoM8xPkZJ9Qe:-Kek2Cx4uEhLjnLzUkD8:b-gqugdy) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L20)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

Does Solidity support '1e' notation? I can't find it anywhere in the docs
</blockquote></details>

Not that's the best answer I could've hoped for (I did not think about this option)
Related: https://github.com/ethereum/solidity/issues/1671

---

*[contracts/Trustcoin.sol, line 24 at r3](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130#-KejDwfPjnz_PTi7q3bx:-Kek4fSLMtSBP_YU1g3J:b-xobxk6) ([raw file](https://github.com/wetrustplatform/rosca-contracts/blob/3997eff2117e0efc5803a543c4edfddce13cb7c6/contracts/Trustcoin.sol#L24)):*
<details><summary><i>Previously, morelazers (Tom Nash) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Actually, I see that we also have `var`s in the contract (probably originating from the source contracts). Can we change them to their explicit types? this is more dangerous than the `uint` type definition as in the famous example of loop variable (you might have seen it):
`for (var i = 0; i < 1000; i++) { /* do something; */  }`
would run forever because the compiler would decide 8 bits are enough for `i`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/wetrustplatform/rosca-contracts/130)*

rmerom (commented):
I think we're almost ready to go. Just need to move to a new repo (see slack) and move the migrate methods to be invoked from the old contract

morelazers:
Moved to a new repo https://github.com/WeTrustPlatform/token-contracts

PR #133, opened 2017-03-07
fix breaking Points , nothing has been upgraded yet

shine2lay:
the code should now work as before with truffle 3.0, non of the recommended standard upgrade has been added

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/133)

### DIFF
--- a/contracts/tokenableROSCA.sol
+++ b/contracts/tokenableROSCA.sol
@@ -11,7 +11,7 @@ pragma solidity ^0.4.4;
  * The discount (gap between bid and total round contributions) is dispersed
  * evenly between the participants.
  */
-contract ROSCA {
+contract tokenableROSCA {
 
   ////////////
   // CONSTANTS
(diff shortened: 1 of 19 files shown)


rmerom:
Just to better understand: should this be working under truffle 3, but no longer under truffle 2?

shine2lay:
i have not tried in truffle 2

rmerom:
ok cool, let's just get everything under control in truffle 2 (the 2 remaining PRs), then proceed with this.
you might also want to update this branch with #134 to see it helps travis.
i'll meanwhile read about the upgrade and review this too.

rmerom:
were you able to build on truffle 3 with this version?
(i'm asking because I can't see `DefaultBuilder` or one of the other builders in `truffle.js`)

shine2lay:
that's only if you use ```truffle build``` , i don't think we use that command at all. I can add the Default builder if you would like me to

shine2lay:
will PR later on a different branch

PR #145, opened 2017-03-16
dont throw if there are no fees to collect

rmerom:
We should throw only when an action's required preconditions are not met. It's perfectly fine for foreperson to collect 0 fees if they want to (and it's more understandable by FE users if we don't throw).

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/145)


contracts/ROSCA.sol
```diff
@@ -528,9 +528,6 @@ contract ROSCA {
    * does the bookeeping of that round.
    */
   function endOfROSCARetrieveFees() onlyFromForeperson onlyIfRoscaEnded external {
-    if (totalFees == 0) {
```
shine2lay, line 531: just making sure i understand this correctly, now foreperson will be able to call Collect Fees any number of times right? even when the totalFees is 0 ?

shine2lay (approved):

Issue #132, opened 2017-03-07
Upgrade to truffle 3


rmerom:

Asone:
seems like can be closed nope ? as it has been implemented in package.json with the last pull request (see #157)

shine2lay:
@Asone yes, thanks for the reminder

PR #152, opened 2017-04-29
Fix the ways events are tested

shine2lay:
part 2 of truffle upgrade: 
- events are checked through receipt instead of watching 

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/152)


rmerom (approved):
Beautiful!
So much more concise...

rmerom:
#132

PR #169, opened 2017-08-14
Expose dynamic state variable

shine2lay:
some of the dynamic state variable were exposed in the previous commit

---
This change is [Reviewable](https://reviewable.io/reviews/wetrustplatform/rosca-contracts/169)


shine2lay:
included in #170

PR #173, opened 2017-10-13
Change default current round

shine2lay:


test/startRoundUnitTest.js
```diff
@@ -45,7 +46,8 @@ contract('ROSCA startRound Unit Test', function(accounts) {
     }));
 
     it("Throws when calling startRound before roundStartTime (including round = 0)", co(function* () {
```
sihoang: the description and test case need to be updated. Maybe `Throws when calling startRound during round 1`?

test/withdrawUnitTest.js
```diff
@@ -25,10 +25,10 @@ contract('ROSCA withdraw Unit Test', function(accounts) {
 
     it("Throws when calling withdraw from a non-member", co(function* () {
         yield Promise.all([
-            ethRoscaHelper.contribute(0, consts.CONTRIBUTION_SIZE),
-            ethRoscaHelper.contribute(1, consts.CONTRIBUTION_SIZE),
-            ethRoscaHelper.contribute(2, consts.CONTRIBUTION_SIZE),
-            ethRoscaHelper.contribute(3, consts.CONTRIBUTION_SIZE),
+            ethRoscaHelper.contribute(0, 2 * consts.CONTRIBUTION_SIZE),
```
sihoang, line 28: this is not related to Round 0 -> 1?

test/withdrawUnitTest.js
```diff
@@ -44,23 +44,23 @@ contract('ROSCA withdraw Unit Test', function(accounts) {
...
-        const ACTUAL_CONTRIBUTION = consts.CONTRIBUTION_SIZE * 0.8;
+        const ACTUAL_CONTRIBUTION = consts.CONTRIBUTION_SIZE * 1.8;
+        const EXPECTED_WITHDRAWAL = consts.CONTRIBUTION_SIZE * 0.8;
 
         yield roscaHelper.contribute(0, ACTUAL_CONTRIBUTION);
 
         let result = yield roscaHelper.withdraw(0);
         let log = result.logs[0];
 
-        assert.equal(log.args.user, accounts[0], "LogContributionMade doesn't display proper user value");
```
sihoang, line 54: we actually call contribute. Should this assert pass?

shine2lay: We are only watching for LogFundsWithdrawal, it was an error on previous version. only the error string was wrong. the assert is right

test/withdrawUnitTest.js
shine2lay: it is, this test originally test withdrawal in round 0 , but we are in round 1 , we need to contribute more than round 1's contribution to be able to withdraw

test/startRoundUnitTest.js
shine2lay: i think we just need to get rid of `(including round 0)` because it goes through all the rounds

test/startRoundUnitTest.js
```diff
@@ -18,6 +18,14 @@ contract('ROSCA startRound Unit Test', function(accounts) {
       roscaHelper = new ROSCAHelper(accounts, (yield utils.createEthROSCA()));
     }));
 
+  it("checks that round 1 can be longer one roundPeriod", co(function* () {
```
sihoang, line 21: indentation

test/startRoundUnitTest.js
```diff
@@ -18,6 +18,14 @@ contract('ROSCA startRound Unit Test', function(accounts) {
...
+    utils.increaseTime(consts.ROUND_PERIOD_IN_SECS);
+    yield utils.assertThrows(roscaHelper.startRound(),
+      "expected calling startRound before roundStartTime to throw");
+    utils.increaseTime(consts.START_TIME_DELAY);
+    yield roscaHelper.startRound();
```
sihoang, line 26: do we need to assert the startRound succeeds?

sihoang (approved):
merge for code audit. Please submit the fix later.

PR #182, opened 2017-12-27
Address audit #2 + tests , refactor cleanUpPreviousRound

### LINKED ISSUE #184: Audit result - Major Flaws
- [x] - Line 206: the loop i < members_.length will never end in a case when members_.length >= 2^16.
- [x] - Line 207: while zero member address plays special role in the code of this smart contract, nothing prevents the deployer of this contract from adding zero address as one of member which would break the contract logic.
- [x] - Line 302: due to the accumulated errors in totalDiscounts >= it may evaluate to false even if the member already came out of the debt.
- [x] - Line 473: in case of using totalDiscounts, due to the accumulated errors, members will be able to withdraw less than they should be able to.
- [x] - Line 609: smart contract will become self-destroyed even if tokenContract.transfer(foreperson, balance) transfer has failed, making tokens lost forever.

### LINKED ISSUE #185: Audit result - Arithmetic Overflow Issues
- [x] Line 19: the contract ROSCA uses the mixture of integer types of different bit-withdraws. This makes the controlling of the potential overflows harder because Solidity usually truncates a result of arithmetic operation to withdraw of a wider parameter. So, this will overflow:
— uint8 a = 100
— uint8 b = 200
— uint16 c = a + b; // c is now 44, not 300 as one may expect

- [x] Line 193: underflow is possible

shine2lay:
- Line 206: the loop i < members_.length will never end in a case when members_.length >= 2^16.
- Line 207: while zero member address plays special role in the code of this smart contract, nothing prevents the deployer of this contract from adding zero address as one of member which would break the contract logic.
- Line 609: smart contract will become self-destroyed even if tokenContract.transfer(foreperson, balance) transfer has failed, making tokens lost forever.
- Line 193: underflow is possible.

Also refactored cleanUpPreviousRound

partially closes #185 
partially closes #184

### DIFF
--- a/contracts/ROSCA.sol
+++ b/contracts/ROSCA.sol
@@ -197,7 +197,7 @@ contract ROSCA {
     require(roundPeriodInSecs_ != 0 &&
       startTime_ >= SafeMath.sub(now, MAXIMUM_TIME_PAST_SINCE_ROSCA_START_SECS) &&
       serviceFeeInThousandths_ <= MAX_FEE_IN_THOUSANDTHS &&
-      members_.length > 1 && members_.length <= 256); // min member count is 1 and max member count is 256
+      members_.length > 1 && members_.length <= 256); // member count must be 1 < x < 256
 
     roundPeriodInSecs = roundPeriodInSecs_;
     contributionSize = contributionSize_;
--- a/contracts/ROSCA.sol
+++ b/contracts/ROSCA.sol
@@ -127,6 +127,11 @@ contract ROSCA {
     reentrancyLock = false;
   }
 
+  modifier onlyNonZeroAddress(address toCheck) {
+    require(toCheck != address(0));
+    _;
+  }
+
   modifier onlyFromMember {
     require(members[msg.sender].alive);
     _;
@@ -190,9 +195,9 @@ contract ROSCA {
       address[] members_,
       uint16 serviceFeeInThousandths_) {
     require(roundPeriodInSecs_ != 0 &&
-      startTime_ >= (now - MAXIMUM_TIME_PAST_SINCE_ROSCA_START_SECS) &&
+      startTime_ >= SafeMath.sub(now, MAXIMUM_TIME_PAST_SINCE_ROSCA_START_SECS) &&
       serviceFeeInThousandths_ <= MAX_FEE_IN_THOUSANDTHS &&
-      members_.length > 0);
+      members_.length > 1 && members_.length <= 256); // min member count is 1 and max member count is 256
 
     roundPeriodInSecs = roundPeriodInSecs_;
     contributionSize = contributionSize_;
@@ -203,14 +208,15 @@ contract ROSCA {
 
     foreperson = msg.sender;
 
-    for (uint16 i = 0; i < members_.length; i++) {
+    for (uint8 i = 0; i < members_.length; i++) {
       addMember(members_[i]);
     }
 
     LogStartOfRound(currentRound);
   }
 
-  function addMember(address newMember) internal {
+  function addMember(address newMember) onlyNonZeroAddress(newMember) internal {
+
     require(!members[newMember].alive);  // already registered
 
     members[newMember] = User({paid: false , credit: 0, alive: true, debt: false});
@@ -245,7 +251,7 @@ contract ROSCA {
    * user with winnings
    */
   function cleanUpPreviousRound() internal {
-    uint256 winnerIndex;
+    uint256 winnerIndex = membersAddresses.length + 1; // to avoid default (0) to be winner
     bool winnerSelectedThroughBid = (winnerAddress != 0);
     uint16 numUnpaidParticipants = uint16(membersAddresses.length) - (currentRound - 1);
     // for pre-ordered ROSCA, pick the next person in the list (delinquent or not)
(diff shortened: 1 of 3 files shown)


contracts/ROSCA.sol
```diff
@@ -600,15 +606,18 @@ contract ROSCA {
...
-    LogEmergencyWithdrawalPerformed(getBalance(), currentRound);
+	  LogEmergencyWithdrawalPerformed(getBalance(), currentRound);
+    bool failedSendingFunds = false;
     // Send everything, including potential fees, to foreperson to disperse offline to participants.
     bool isEthRosca = (tokenContract == address(0));
     if (!isEthRosca) {
       uint256 balance = tokenContract.balanceOf(address(this));
-      // we don't care much about the success of transfer` here as there's not much we can do.
-      tokenContract.transfer(foreperson, balance);
+	    failedSendingFunds = tokenContract.transfer(foreperson, balance);
```
sihoang: indentation

contracts/ROSCA.sol
```diff
@@ -600,15 +606,18 @@ contract ROSCA {
    * this sends all the funds to the foreperson by selfdestructing this contract.
    */
   function emergencyWithdrawal() onlyFromForeperson onlyIfEscapeHatchActive external {
-    LogEmergencyWithdrawalPerformed(getBalance(), currentRound);
+	  LogEmergencyWithdrawalPerformed(getBalance(), currentRound);
```
sihoang: indentation

contracts/ROSCA.sol
sihoang: the name `failedSendingFunds` is misleading

contracts/ROSCA.sol
```diff
@@ -600,15 +606,18 @@ contract ROSCA {
...
+    bool failedSendingFunds = false;
```
sihoang: please use different var name. It's misleading because `failedSendingFunds = tokenContract.transfer(foreperson, balance);` means the opposite

contracts/ROSCA.sol
```diff
@@ -600,15 +606,18 @@ contract ROSCA {
...
     // Send everything, including potential fees, to foreperson to disperse offline to participants.
     bool isEthRosca = (tokenContract == address(0));
     if (!isEthRosca) {
```
sihoang: if isEthRosca is true?

contracts/ROSCA.sol
```diff
@@ -190,9 +195,9 @@ contract ROSCA {
       address[] members_,
       uint16 serviceFeeInThousandths_) {
     require(roundPeriodInSecs_ != 0 &&
-      startTime_ >= (now - MAXIMUM_TIME_PAST_SINCE_ROSCA_START_SECS) &&
+      startTime_ >= SafeMath.sub(now, MAXIMUM_TIME_PAST_SINCE_ROSCA_START_SECS) &&
       serviceFeeInThousandths_ <= MAX_FEE_IN_THOUSANDTHS &&
-      members_.length > 0);
+      members_.length > 1 && members_.length <= 256); // member count must be 1 < x < 256
```
sihoang: < 256 or <= 256? the code and comment are different

contracts/ROSCA.sol
```diff
@@ -245,7 +251,7 @@ contract ROSCA {
    * user with winnings
    */
   function cleanUpPreviousRound() internal {
-    uint256 winnerIndex;
+    uint256 winnerIndex = membersAddresses.length + 1; // to avoid default (0) to be winner
```
sihoang: I think it's better to have default winner rather than potentially no winner at all. Default of 0 is fine.

FYI, if you insist on avoiding default 0, it's better to set the value of winnerIndex in findSemiRandomWinner because winnerIndex is only effective if winnderAdress == 0 (i.e winnerSelectedThroughBid is false)

sihoang: also, after reading the clean code, please refactor cleanUpPreviousRound and swapWinner. Their logic is dangled. Specially `swapWinner` looks bad because the winnerIndex is passed in but only used when winnerSelectedThroughBid is false.

Why we introduce another variable `winnerSelectedThroughBid` when we have typesOfROSCA.AUCTION_ROSCA?

shine2lay: i am confused with the first statement about `winnerIndex` only being used when `winnerSelectedThroughBid` is false, if you look at https://github.com/shine2lay/rosca-contracts/blob/2fcd2521fa35d06ffaa62bd1f813d4c87051c4e4/contracts/ROSCA.sol#L370. you can see it winnerIndex is being used regardless of the `winnerSelectedThroughBid`

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: this is correct, we only want to call `transfer` method of the token contract, if the rosca is a token rosca

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
shine2lay: reverted

shine2lay: that being said, i think you are absolutely right in that there were alot tangled logic, i did some refactoring to clean it up

shine2lay:
@sihoang PTAL, i refactored the cleanUpPreviousRound

sihoang:
looks much better!

I am just wondering in the `emergencyWithdrawal` if it's ethRosca, do we need to move the ETH out?

shine2lay:
@sihoang selfdestruct() by default send all the funds in the contract to the destination before destroying the contract.

sihoang (approved):

PR #183, opened 2017-12-29
Address Audit #3 + fix tests + refactor userTotalCredit

### LINKED ISSUE #184: Audit result - Major Flaws

shine2lay:
- Line 302: due to the accumulated errors in totalDiscounts >= it may evaluate to false even if the member already came out of the debt.
- Line 473: in case of using totalDiscounts, due to the accumulated errors, members will be able to withdraw less than they should be able to.

close #184

### DIFF
--- a/contracts/ROSCA.sol
+++ b/contracts/ROSCA.sol
@@ -275,7 +275,7 @@ contract ROSCA {
     }
     uint256 currentRoundTotalDiscounts = removeFees(potSize() - lowestBid);
     uint256 roundDiscount = currentRoundTotalDiscounts / membersAddresses.length;
-    totalDiscounts += roundDiscount;
+    totalDiscounts += currentRoundTotalDiscounts;
     members[winnerAddress].credit += removeFees(lowestBid);
     members[winnerAddress].paid = true;
     LogRoundFundsReleased(winnerAddress, lowestBid, roundDiscount, currentRound);
@@ -299,7 +299,7 @@ contract ROSCA {
       address candidate = membersAddresses[index];
       if (!members[candidate].paid) {
         winnerIndex = index;
-        if (members[candidate].credit + totalDiscounts >= requiredContribution()) {
+        if (members[candidate].credit + userDiscounts() >= requiredContribution()) {
           // We found a non-delinquent winner.
           winnerAddress = candidate;
           break;
@@ -336,8 +336,8 @@ contract ROSCA {
         // they could not bid), to correctly calculate their delinquency.
         debit = SafeMath.add(debit, removeFees(potSize()));
       }
-      if (credit + totalDiscounts < debit) {
-        grossTotalFees = SafeMath.sub(grossTotalFees, (debit - credit - totalDiscounts));
+      if (credit + userDiscounts() < debit) {
+        grossTotalFees = SafeMath.sub(grossTotalFees, (debit - credit - userDiscounts()));
       }
     }
 
@@ -408,7 +408,7 @@ contract ROSCA {
       // so we check whether their credit w/o that winning is non-delinquent.
       // check that credit must defaultPot (when debt is set to true, defaultPot was added to credit as winnings) +
       // currentRound in order to be out of debt
-      if (SafeMath.sub(member.credit + totalDiscounts, removeFees(potSize())) >= requiredContribution()) {
+      if (SafeMath.sub(member.credit + userDiscounts(), removeFees(potSize())) >= requiredContribution()) {
           member.debt = false;
       }
     }
@@ -429,7 +429,7 @@ contract ROSCA {
     require(!members[msg.sender].paid  &&
         currentRound != 0 &&  // ROSCA hasn't started yet
         // participant not in good standing
-        members[msg.sender].credit + totalDiscounts >= requiredContribution() &&
+        members[msg.sender].credit + userDiscounts() >= requiredContribution() &&
         // bid is less than minimum allowed
         bid >= SafeMath.mul(potSize(), MIN_DISTRIBUTION_PERCENT) / 100);
 
@@ -470,7 +470,7 @@ contract ROSCA {
   function withdraw() onlyFromMember onlyIfEscapeHatchInactive nonReentrant external returns(bool success) {
     require (!members[msg.sender].debt || endOfROSCA); // delinquent winners need to first pay their debt
 
-    uint256 totalCredit = members[msg.sender].credit + totalDiscounts;
+    uint256 totalCredit = members[msg.sender].credit + userDiscounts();
 
     uint256 totalDebit = members[msg.sender].debt
         ? removeFees(potSize())  // this must be end of rosca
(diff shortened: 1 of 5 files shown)


contracts/ROSCA.sol
```diff
@@ -615,6 +615,17 @@ contract ROSCA {
 	// HELPER FUNCTIONS
 	////////////////////
 
+
+	/**
+	 * @dev calculates the user's discount amount from the total discount
+	 * @return uint256
+	 */
+	function userTotalCredit(address memberAddress) internal constant returns (uint256) {
+    uint256 userDiscount = totalDiscounts / membersAddresses.length;
```
sihoang: indentation

contracts/ROSCA.sol
```diff
@@ -615,6 +615,17 @@ contract ROSCA {
 	// HELPER FUNCTIONS
 	////////////////////
 
+
+	/**
+	 * @dev calculates the user's discount amount from the total discount
+	 * @return uint256
+	 */
+	function userTotalCredit(address memberAddress) internal constant returns (uint256) {
```
sihoang: indentation

contracts/ROSCA.sol
```diff
@@ -502,7 +502,7 @@ contract ROSCA {
    * @return int256
    */
   function getParticipantBalance(address user) onlyFromMember external constant returns(int256) {
-    int256 totalCredit = int256(members[user].credit + totalDiscounts);
+    int256 totalCredit = int256(userTotalCredit(user));
```
sihoang, line 505: no need to cast to int256

sihoang (commented):
looks good so far. Mergeable.

contracts/ROSCA.sol
shine2lay: you have to cast it, `userTotalCredit` returns a uint256, solidity doesn't allow implicit cast between uint256 to int256. if you don't have this cast, you get this error 
image

contracts/ROSCA.sol

contracts/ROSCA.sol

contracts/ROSCA.sol
```diff
@@ -611,9 +611,20 @@ contract ROSCA {
...
+
+
+  /**
+   * @dev calculates the user's discount amount from the total discount
+   * @return uint256
+   */
+  function userTotalCredit(address memberAddress) internal constant returns (uint256) {
+    uint256 userDiscount = totalDiscounts / membersAddresses.length;
+
+  	return SafeMath.add(members[memberAddress].credit, userDiscount);
```
knd: Why's there an extra indentation here?

contracts/ROSCA.sol
```diff
@@ -275,7 +275,7 @@ contract ROSCA {
     }
     uint256 currentRoundTotalDiscounts = removeFees(potSize() - lowestBid);
     uint256 roundDiscount = currentRoundTotalDiscounts / membersAddresses.length;
```
knd, line 277: Why do we need to know `roundDiscount` and log it? Why not:

```
LogRoundFundsReleased(winnerAddress, lowestBid, currentRoundTotalDiscounts, currentRound);
```

knd:
Could you explain the comment on this line? 

https://github.com/WeTrustPlatform/rosca-contracts/pull/183/files#diff-ac2158c046f58e4fb91fe99a56e138c2R294

> // Note this randomness does not require high security, that's why we feel ok with using the block's timestamp.

What's the security concern behind this? We should treat smart contract as highly secure entity. It could be either secure OR insecure and we need to be aware of it.

contracts/ROSCA.sol
```diff
@@ -299,7 +299,7 @@ contract ROSCA {
       address candidate = membersAddresses[index];
       if (!members[candidate].paid) {
         winnerIndex = index;
-        if (members[candidate].credit + totalDiscounts >= requiredContribution()) {
+        if (userTotalCredit(candidate) >= requiredContribution()) {
```
knd, line 302: I thought we'd like to check something like: 

```
members[memberAddress].credit >= requiredContribution()
```
the amount user has contributed so far against what's the amount he/she's supposed to contributed up to current round? 

Why the user credit has to have surplus (which is: totalDiscounts / membersAddresses.length) ?

contracts/ROSCA.sol
```diff
@@ -85,7 +85,7 @@ contract ROSCA {
   bool public endOfROSCA = false;
   bool public forepersonSurplusCollected = false;
   // A discount is the difference between a winning bid and the pot value. totalDiscounts is the amount
-  // of discounts accumulated so far, divided by the number of ROSCA participants.
+  // of discounts accumulated so far
```
knd, line 88: The calculation and probably the meaning of `totalDiscounts` is are changed after this pull request. 

Do we want to change the variable name too? 

If `totalDiscounts` name is still the intention, then the name is totally inappropriate in the past.

contracts/ROSCA.sol
shine2lay: its not a `need`, more like a `want`. It is helpful to log out more information and it would also take less computation on client side to show roundDiscount.

contracts/ROSCA.sol
shine2lay: traditional ROSCA have this logic where if you are supposed to pay 1 ETH per round, and you have 0.1 ETH discount (earned from the last round). you only need to pay up 0.9 ETH this round.

contracts/ROSCA.sol
shine2lay: there is not a better name for this variable IMHO.

contracts/ROSCA.sol
shine2lay: should be fixed now, it was caused by my IDE

shine2lay:
re Randomness Security: we don't care that user maybe able to manipulate the randomness (i.e the variable 'now' which represent the block's timestamp), because the attacker needs to be 1). a member of the ROSCA, 2). haven't won a round before
in order to win the round. so even if the attacker manipulate (which will take a lot of effort because you need to create the block in order to manipulate) the randomness, he/she can only win once. no more

contracts/ROSCA.sol
sihoang: I think it's better to remove the roundDiscount. We don't need to save computation on client side by putting it on a smart contract.
Please take care of non-backward compatible change in the client

sihoang: actually, there's risk of non-backward compatible change. I am fine with leaving as is now. @Kivutar @knd let me know if you think we have enough time to test on the client changes?

contracts/ROSCA.sol
sihoang: why do we need to use both int256 and uint256?

shine2lay: Because the function will return negative value if user    owe the contract by the amount they owe. This is for read only.

contracts/ROSCA.sol
shine2lay: i think we have to remove it, because that calculation is no longer accurate, consider this scenario 
1/3 (round 1 discount) + 1/3 (round 2 discount) + 1/3 (round 3 discount) != 3/3 (totalDiscount)
0.999999 != 1

knd (approved):

